PR #36, opened 2013-03-10
JENKINS-17009  Parameters with newlines should be stored as TextParameterValue 

centic9:
Simple fix for problems with multiline parameters: if the value of the parameter contains a newline character then use a TextParameterValue, as this handles newlines better than StringParameterValue.

The result is that downstream plugins like the rebuild plugin can correctly handle these kind of parameter values then.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.424</version>
+        <version>1.478</version>
     </parent>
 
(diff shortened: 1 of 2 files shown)


zalan-axis:
Its late 2015 and this problem is still there is Jenkins 1.632 + Parameterized Trigger plugin 2.29.

I think this is a no-brainer...
With the "Current build parameters", you have knowledge of the exact hudson.model.ParameterValue type. In case of "Predefined parameters" and "Parameters from property files" you have to make an educated guess. And TextParameterValue sure seems like a better fit if the string contains newlines than StringParameterValue does. 

Can't you be pragmatic about this?

PR #62, opened 2014-02-27
Attempt to Convert StringParameterValues to the type defined in the project

zeroem:
Rather than just passing StringParameterValues to the downstream job, attempt to convert them to the downstream project's defined type, easing integration with parameters that have specialized ParameterValue objects (eg. NodeLabelParameter).

The basic logic goes:

If the trigger parameter is an instance of `StringParameterValue` and the downstream parameter definition is an instance of `SimpleParameterDefinition`, replace the parameter value with:

`downstreamParameterDefintion.createValue(triggerParameterValue.value)`


Cigizmoond-Vyhuholev:
Hello!
We have an issue with triggeting build with parameter of class VersionParameterDefinition

PR #78, opened 2014-12-11
[FIXED JENKINS-26031] Added export of triggered builds to XML/JSON API

abayer:


ikedam:
I don't think this work as expected.

wolfs:
Replaced by #82.

mthak:
How do i get the same data using XML . I don't see it exposed via api/xml, i do see it in json though

wolfs:
<jenkins-url>/job/<job-name>/<build-id>/api/xml?depth=1 works like a charm for me.

PR #79, opened 2014-12-17
Provide an option to skip triggering when a build job is invoked manually

kevinvandervlist:
Add an option which allows the user to choose between executing triggers like they normally do, or to skip them when a user was the cause of the build of this job. This allows users to decide whether or not they want to invoke a whole pipeline of jobs or if they just want to invoke a single job.


oleg-nenashev:
+1 for having this feature in Conditional Build Step.
Causes are partially supported there, so I doubt it makes sense to merge this pull request.

kevinvandervlist:
Sure, that's fine with me. We only needed it as a temporary solution anyway, we don't even use it anymore.

I'll just close the PR then.

oleg-nenashev:
@kevinvandervlist 
Thanks a lot for your efforts in any case

PR #82, opened 2015-03-03
[FIXED JENKINS-26031] Added export of triggered builds and jobs to XML/JSON API

wolfs:
Replaces #78.
To try it out go to <build-url>/api/json?pretty=true&depth=1 of a build that triggered another build via the parameterized trigger build step. Then you will see something like this:
```json
{
  "actions" : [
    ...,
    {
      "triggeredBuilds" : [
        
      ],
      "triggeredProjects" : [
        {
          "name" : "chain-step-1",
          "url" : "http://localhost:8080/job/chain-step-1/",
          "color" : "blue"
        }
      ]
    }
  ],
...
}
```

PR #83, opened 2015-03-05
Update core to 1.580.1 - Require Java 1.6

wolfs:
Releasing did not work for me with maven 3 - updating core version.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameterFactory.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameterFactory.java
@@ -7,35 +7,21 @@ package hudson.plugins.parameterizedtrigger;
 import com.google.common.collect.Lists;
 import hudson.Extension;
 import hudson.FilePath;
-import hudson.FilePath.FileCallable;
 import hudson.Util;
 import hudson.model.AbstractBuild;
 import hudson.model.TaskListener;
-import hudson.remoting.VirtualChannel;
 import hudson.util.FormValidation;
-import hudson.util.LogTaskListener;
-import hudson.util.StreamTaskListener;
-import hudson.util.VariableResolver;
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileFilter;
-import java.io.FileNotFoundException;
-import java.io.FileReader;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import org.apache.commons.lang.StringUtils;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
 
 import java.io.IOException;
-import java.io.PrintStream;
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
 import java.nio.charset.UnsupportedCharsetException;
-import java.util.ArrayList;
 import java.util.List;
-import org.apache.commons.lang.StringUtils;
-import org.apache.tools.ant.DirectoryScanner;
-import org.apache.tools.ant.types.FileSet;
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 
 /**
@@ -160,9 +146,6 @@ public class FileBuildParameterFactory extends AbstractBuildParameterFactory {
                     return FormValidation.error("Unsupported Encoding");
                 } catch(IllegalCharsetNameException e) {
                     return FormValidation.error("Bad Encoding Name");
-                }
-                if(!ParameterizedTriggerUtils.isSupportNonAsciiPropertiesFile()) {
-                    return FormValidation.warning("Non-ascii properties files are supported only since Java 1.6.");
                 }
             }
             return FormValidation.ok();
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.480</version>
+        <version>1.580.1</version>
     </parent>
     
     <licenses>
@@ -41,6 +41,12 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-project</artifactId>
+            <version>1.4</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>promoted-builds</artifactId>
             <version>2.10</version>
             <optional>true</optional>
@@ -69,13 +75,6 @@
         <version>1.3.1</version>
         <optional>true</optional>
       </dependency>      
-      <dependency>
-        <!-- Used for @IgnoreJRERequirement -->
-        <groupId>org.jvnet</groupId>
-        <artifactId>animal-sniffer-annotation</artifactId>
-        <version>1.0</version>
-        <optional>true</optional>
-      </dependency>
     </dependencies>
 
     <build>
@@ -99,7 +98,7 @@
               <configuration>
                 <signature>
                   <groupId>org.jvnet.animal-sniffer</groupId>
-                  <artifactId>java1.5</artifactId>
+                  <artifactId>java1.6</artifactId>
                   <version>1.0</version>
                 </signature>
               </configuration>
--- a/src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameters.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameters.java
@@ -1,17 +1,30 @@
 package hudson.plugins.parameterizedtrigger;
 
+import com.google.common.base.Predicate;
+import com.google.common.collect.Collections2;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Util;
-import hudson.model.*;
-import hudson.util.FormValidation;
 import hudson.matrix.AxisList;
 import hudson.matrix.Combination;
 import hudson.matrix.MatrixBuild;
 import hudson.matrix.MatrixProject;
 import hudson.matrix.MatrixRun;
+import hudson.model.AbstractBuild;
+import hudson.model.Action;
+import hudson.model.Descriptor;
+import hudson.model.ParameterValue;
+import hudson.model.ParametersAction;
+import hudson.model.Run;
+import hudson.model.StringParameterValue;
+import hudson.model.TaskListener;
+import hudson.util.FormValidation;
+import org.apache.commons.lang.StringUtils;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
 
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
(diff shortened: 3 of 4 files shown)


KostyaSha:
Who allowed you merging this PR?

src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameterFactory.java
```diff
@@ -161,9 +147,6 @@ public FormValidation doCheckEncoding(@QueryParameter String encoding) {
                 } catch(IllegalCharsetNameException e) {
                     return FormValidation.error("Bad Encoding Name");
                 }
-                if(!ParameterizedTriggerUtils.isSupportNonAsciiPropertiesFile()) {
-                    return FormValidation.warning("Non-ascii properties files are supported only since Java 1.6.");
-                }
```
jglick: Why??

jglick: Seems to be reverting parts of #55, I guess because of the assumption that Java 6 is required, though not quite complete since the Java 5-related text in `help-encoding.html` was left.

jglick:
Is @ikedam no longer maintaining this plugin?

In general, updating the core baseline for a reason is fine—even if the reason is simply that testing and supporting old versions has become onerous—but you should not update it just because you did not know you needed to run

    mvn -B org.apache.maven.plugins:maven-release-plugin:2.5.1:prepare org.apache.maven.plugins:maven-release-plugin:2.5.1:perform

to work around a bug in the Git provider for the Maven release plugin.

ikedam:
I no longer maintain parameterized-trigger as other developers started to maintain it.

daniel-beck:
@wolfs Could you please clarify why these changes were implemented? The PR doesn't mention rationales for these changes, and neither has an obvious reason.

* The necessary workarounds for 1.5 compatibility, assuming it wasn't broken anyway otherwise, don't seem to be extensive enough to impact development and could just be left in: One form validation, one try/catch.
* It's not clear what the release issue is. I don't see how it could be the one Jesse refers to (which is also mentioned [in the wiki](https://wiki.jenkins-ci.org/display/JENKINS/Hosting+Plugins#HostingPlugins-Workingaroundcommonissues)) as it hasn't yet been fixed in the parent POM in 1.580.1, so more information would be helpful to determine whether this was actually necessary.

wolfs:
OK, so here is my story:
I needed more information from the REST API, so I extended #78 to #82. I waited a little if I get some feedback on the changes and since there was not any I decided - the changes being minor - to merge #82 and pull a release myself.
In the process of building the release I ran into an issue with building the plugin with a newer version of maven and was not able to build it. After updating to a new core the build worked.  Therefore, I figured it would be a good idea to go to this new core version.
After updating to the new core version I saw a warning about the code handling the Java 5 compatibility. Since Java 5 support has been dropped from Jenkins Core a while ago I also removed the corresponding code. I then built the release and updated the wiki site.
I did not try to contact the maintainers because it was not clear to me who they were at the time. Looking at git history didn't give me much information. Also, given that #83 was a minor change, I needed it in production and I already contributed a much more non-trivial extension to the plugin I figured it would not be necessary - in retrospect this could have been the wrong decision. Who is the current maintainer to contact?
I am sorry if I caused any problems. Could you please tell me if there is any issue with this commit? Have there been any problems? If someone wants to discuss this I would prefer to do this via another medium which is better suited for this like Skype or Google Hangouts.

KostyaSha:
1) Developers: https://github.com/jenkinsci/parameterized-trigger-plugin/blob/master/pom.xml#L24-L33
2) Have you wrote to developers maillist? 
3) Still see no good description why core was updated. I can't use new plugin version on 1.565.3 for example.
Bumping core to the latest without appropriate description is bad practise in jenkins development.

daniel-beck:
@wolfs Thanks for the explanation! What JDK and Maven versions were you using to build this plugin?

@KostyaSha How about you revert the parent POM version again locally and try to build the plugin to determine what Stefan's build problem was, and how to resolve it? It's not like this cannot be reduced again in a subsequent release.

wolfs:
@daniel-beck Mvn: 3.2.2, JDK: 1.7.0_45

wolfs:
@KostyaSha Could you please tell me why updating core version to the previous released LTS branch is a bad practice? After my understanding this is actually best practice. If you could tell me where this is documented I would be happy to know.

KostyaSha:
@daniel-beck i'm not a maintainer of this plugin and i have no time.
@wolfs have you checked how many active installations does it have? Because everybody trying to have 1y of compatibility.

wolfs:
@KostyaSha Could you please point me to the documentation of the policy to have 1 year of compatibility?
If I understand correctly the problem you have with this change is the new required core version?
You can still install this plugin on an older jenkins version and it will work - you only need to ignore the warning. Or install an older version of the plugin - or even do not update at all.
@daniel-beck Can you give me a hint how to do a release build without releasing the plugin again? Then I could try to reproduce the problem.

daniel-beck:
> Could you please point me to the documentation of the policy to have 1 year of compatibility?

There is none that I know of. The 'compatibility' project philosophy seems to be limited to data compatibility. Maybe the CloudBees devs do that because that's their LTS support duration, but the archetype is already on 1.580.x for new plugins. No one-year compatibility there.

In general though, we should strive to not add unnecessarily limit availability of plugins/newer versions. So only increase core requirement if there's a good reason, and maybe even maintain a lower-core branch if you're going to target a very recent release (like Subversion Plugin did with 2.4.x). The more used a plugin is, the more important this becomes.

> Can you give me a hint how to do a release build without releasing the plugin again?

Good question. maven-release-plugin docs mention `release:stage` and `release:prepare -DdryRun=true` to perform a modified release. Alternatively, just call it `-alpha-` to release to the experimental update center?

KostyaSha:
@daniel-beck How CloudBees enforced everybody to use 1.580 is other story.  In the same time CloudBees decreases core versions: example https://github.com/jenkinsci/naginator-plugin/pull/15 .

1y (-4 lts versions?) is unspoken rule for all maintainers that supports real jenkins production installations. Of course if you work only with one plugin and localhost installation, then you want to play with recent version. If plugin developers and maintainers approved @wolfs maintaining - then i'm ok with such changes, because it will be his decision.
I also remember that some maintainers replied about 2y support (but imho it unnecessary).

@daniel-beck afair stage operation do -rc and it designed to work with rollback/release. 

@wolfs It maybe useful to read maven-release-plugin documentation if you have problems with this plugin. You can try http://maven.apache.org/maven-release/maven-release-plugin/prepare-mojo.html#dryRun but more deep variant to check scm operations (that release-plugin < 2.5 has) will be with releasing to experimental update center.

daniel-beck:
@KostyaSha While I agree with your general argument that core requirements should not be increased abitrarily, let me respond to the following specific things you mention:

> unspoken rule for all maintainers that supports real jenkins production installations

It's hardly reasonable to expect everyone to be aware of such a rule if it's *unspoken* and only known to those who support *real* installations (whatever that is). If your expectation is that this rule is followed, make sure it is well known to *everyone*, as well as widely accepted by providing good arguments for it.

And if a plugin needs a new core feature, such as an extension point, it absolutely makes sense to raise the parent POM version to allow it to be used. It's not like the previous versions cannot be used anymore, and important fixes can always be backported. Again, Subversion Plugin 2.4.x worked pretty well.

>  I also remember that some maintainers replied about 2y support (but imho it unnecessary).

Was there a poll of some kind on the dev list, or what are you referring to?

> How CloudBees enforced everybody to use 1.580 is other story.

Not sure what you mean. AFAICT they mostly keep the core requirements low because of the support duration, at least on their own plugins. I think they're not on 1.580 anywhere, except when absolutely necessary (e.g. Workflow; also a great example for a plugin needing recent core for a very good reason).

Regarding the plugin archetype being updated rather aggressively, one possible reason I can think of is to allow new devs to use all core features (such as extension points) and fixes out of the box. New plugins take a while to get wider distribution anyway. But it still illustrates that "the rule" is hardly universal.

kohsuke:
I'm saddened to see that accusational tones of earlier comments to Stefan.

He wanted to scratch some itches, he waited and saw no actions, so he took the issue to himself to release it. And there was some release issues, and so he solved it by bumping up the core requirement. I just don't see anything wrong in this. I just don't understand this notion that somehow he has gone against the rules/conventions, or that he should have known better.

A long time contributor like him deserves more respect. And an insult like somehow he is not a maintainer "that supports real jenkins production installations" is just plain unacceptable.

As far as I'm concerned, @wolfs doesn't owe anyone any further changes with regard to this. If somebody else wants to keep the new versions of this plugin compatible with an earlier version of the core than 1.580, then that somebody can go ahead and make more changes to `pom.xml`.

KostyaSha:
@kohsuke this all fine if flow described in this document https://wiki.jenkins-ci.org/display/JENKINS/Governance+Document#GovernanceDocument-Makingchangestoexistingplugins was followed. 
> I did not try to contact the maintainers because it was not clear to me who they were at the time.
And i really interested why some core version can be bad for plugin. Still think that we need documentation page that describes core version changes in plugin developers context.

KostyaSha:
Also, many requests to maillists getting their reviewes.

src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameterFactory.java
wolfs: Thanks for pointing out the missing help text - I will fix this on master.

wolfs:
@kohsuke Thank you for your support.

@KostyaSha Do you have an issue in production with this change? Do you know someone who has an issue? If this is the case then we can easily go forward in trying to solve that problem. Please open a jira ticket for that.

Who is the current maintainer of this plugin? Or is it orphaned now? If the current maintainer suggests to use an older core version I am willing to release the plugin again.

KostyaSha:
@wolfs i have no 1.580 in production and i can't see this new version in UC. I come to this commits because i'm debugging deserialisation issues in buildpipeline-plugin that fails in parametrised actions (probably unrelated). 

2 times when people asked me to fix plugins where maintainer is not active i sent emails to maillist  and waited 1 week for replies, only then i merged and released plugins. In this plugin developers are https://github.com/jenkinsci/parameterized-trigger-plugin/blob/master/pom.xml#L24-L33 

For example i and other person still waiting replies for port-allocator-plugin https://groups.google.com/forum/#!searchin/jenkinsci-dev/port/jenkinsci-dev/I9RIMAHqqTg/yTl248ScHKgJ

jglick:
`<developers>` sections in POMs are in practice useless and should be ignored/deleted. There is no reliable way of telling who thinks of themself as the current maintainer of a plugin. Sometimes it is obvious from the Git history. Otherwise you just need to ask around.

KostyaSha:
@jglick some field mark like `<retired> true</retired>` can help indicate that plugin in state maintainer-wanted.
Workflow can work as:
1) want commit?
2) find devs (1 week delay with notifications to dev list and etc, that i was enforced to do for buildpipeline-plugin)
3) no devs -> mark retired
4) ..ck plugin

KostyaSha:
UPD: i thought it comment to other plugin. But maybe valid to this case also.

jglick:
Might be useful. At any rate I propose closing the discussion here (can use `jenkinsci-dev` if you feel it is productive). If someone wants new releases of this plugin to again be compatible with earlier LTS lines, I suggest just creating a backports branch—it is not hard.

KostyaSha:
@jglick yes, but when i see core bumps:
1) i would like to have a reason for update. This can also help me to provide arguments for updating jenkins core in production.
2) know maybe i need update core in other plugins to exclude some errors.

Thats why i was such :disappointed:

PR #84, opened 2015-05-11
Add Fixed build trigger.

smccarthy:
Adds a Fixed build trigger. A build is considered Fixed if the current build is SUCCESS and the previous build is FAILURE or UNSTABLE.


smccarthy:
I added the JENKINS improvement for this pull request (as stated in the document above) : https://issues.jenkins-ci.org/browse/JENKINS-28354

smccarthy:
@olivergondza Hi. Is there anything else I need to do for this PR? Or is this feature not wanted in the master upstream branch?

Thank you.

olivergondza:
I do not have much experience with this plugin, though the feature makes sense to me.

The change of `ResultCondition` might be unnecessary (as in current form it will break dependent plugins). You should introduce `ResultCondition#isMet(Run<?, ?> run)` that would delegate to the `isMet(Result)` that will not be abstract but will throw `UnsupportedOperationException`. This will not break existing implementation and should give even more flexibility for new conditions via overriding the new method.

olivergondza:
Also, a test is missing.

smccarthy:
Thank you for taking a look. I will make your suggested update and also add a test or two.

PR #85, opened 2015-06-19
[JENKINS-28980] Do not rely on default ArtifactManager implementation

olivergondza:
[JENKINS-28980](https://issues.jenkins-ci.org/browse/JENKINS-28980)


jglick:
Could actually use `compress-artifacts` as a `test`-scoped dependency to validate this if you wanted.

olivergondza:
Yeah, I implemented that as https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/ParameterizedTriggerTest.java#L42. You are right it does not have to be in ATH. I do not feel strongly.

jglick:
Oh OK, good enough.

PR #86, opened 2015-07-17
Enable FindBugs checks on the verify stage

oleg-nenashev:
- [x] Update FindBugs version
- [x] Enable XML report generation
- [x] Enable reporting on https://jenkins.ci.cloudbees.com/job/plugins/job/promotes-builds-plugin/

The code is not clean according to the local run


oleg-nenashev:
@reviewbybees
I'm going to submit fixes as a part of another PR

oleg-nenashev:
YAGNI, Parent POM does it in a better way

PR #87, opened 2015-07-20
[JENKINS-26050] Fix Workflow compatibility for Parameterized Trigger

svanoort:
Per [JENKINS-26050](https://issues.jenkins-ci.org/browse/JENKINS-26050) this fixes workflow compatibility, and adds unit tests for same.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,19 +68,19 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>1.4</version>
+            <version>${workflow.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>1.4</version>
+            <version>${workflow.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
-            <version>1.4</version>
+            <version>${workflow.version}</version>
             <scope>test</scope>
         </dependency>
 
@@ -181,6 +181,7 @@
     <properties>
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <workflow.version>1.4</workflow.version>
     </properties>
 </project>  
 
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
+        <version>1.607</version>
     </parent>
     
     <licenses>
@@ -63,6 +63,19 @@
           <version>1.3</version>
           <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-job</artifactId>
+            <version>1.8</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-cps</artifactId>
+            <version>1.8</version>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
           <groupId>org.mockito</groupId>
           <artifactId>mockito-core</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.609</version>
+        <version>1.580.1</version>
     </parent>
     
     <licenses>
@@ -68,19 +68,19 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>1.8</version>
+            <version>1.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>1.8</version>
+            <version>1.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
-            <version>1.8</version>
+            <version>1.4</version>
             <scope>test</scope>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,8 @@
           <version>1.3</version>
           <scope>test</scope>
         </dependency>
+
+        <!-- Workflow plugin imports are used to test compatibility -->
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
@@ -72,6 +74,12 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
+            <version>1.8</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-basic-steps</artifactId>
             <version>1.8</version>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -181,7 +181,7 @@
     <properties>
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <workflow.version>1.4</workflow.version>
+      <workflow.version>1.4</workflow.version>
     </properties>
 </project>  
 
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.607</version>
+        <version>1.609</version>
     </parent>
     
     <licenses>
--- a/src/main/java/hudson/plugins/parameterizedtrigger/AbstractBuildParameterFactory.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/AbstractBuildParameterFactory.java
@@ -1,9 +1,7 @@
 package hudson.plugins.parameterizedtrigger;
 
 import hudson.ExtensionPoint;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractDescribableImpl;
-import hudson.model.TaskListener;
+import hudson.model.*;
 
 import java.io.IOException;
 import java.util.List;
@@ -24,10 +22,10 @@ public abstract class AbstractBuildParameterFactory extends AbstractDescribableI
      *      The build which the parameterized trigger is configured and executing.
      * @param listener
      *      Connected to the build output.
-     *      
+     *
      * @return can be empty but never null.
      */
-    public abstract List<AbstractBuildParameters> getParameters(AbstractBuild<?,?> build, TaskListener listener)
+    public abstract List<AbstractBuildParameters> getParameters(AbstractBuild<?, ?> build, TaskListener listener)
             throws IOException, InterruptedException, AbstractBuildParameters.DontTriggerException;
 
     @Override
--- a/src/main/java/hudson/plugins/parameterizedtrigger/AbstractBuildParameterFactory.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/AbstractBuildParameterFactory.java
@@ -1,7 +1,9 @@
 package hudson.plugins.parameterizedtrigger;
 
 import hudson.ExtensionPoint;
-import hudson.model.*;
+import hudson.model.AbstractBuild;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.TaskListener;
 
 import java.io.IOException;
 import java.util.List;
@@ -22,10 +24,10 @@ public abstract class AbstractBuildParameterFactory extends AbstractDescribableI
      *      The build which the parameterized trigger is configured and executing.
      * @param listener
      *      Connected to the build output.
-     *
+     *      
      * @return can be empty but never null.
      */
-    public abstract List<AbstractBuildParameters> getParameters(AbstractBuild<?, ?> build, TaskListener listener)
+    public abstract List<AbstractBuildParameters> getParameters(AbstractBuild<?,?> build, TaskListener listener)
             throws IOException, InterruptedException, AbstractBuildParameters.DontTriggerException;
 
     @Override
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
@@ -3,14 +3,9 @@ package hudson.plugins.parameterizedtrigger;
 import com.google.common.collect.ImmutableList;
 import hudson.Extension;
 import hudson.Launcher;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractProject;
-import hudson.model.Action;
-import hudson.model.BuildListener;
+import hudson.model.*;
 import hudson.model.Cause.UpstreamCause;
-import hudson.model.Hudson;
-import hudson.model.Node;
-import hudson.model.Run;
+import jenkins.model.ParameterizedJobMixIn;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import com.google.common.collect.ArrayListMultimap;
@@ -53,14 +48,14 @@ public class BlockableBuildTriggerConfig extends BuildTriggerConfig {
     }
 
     @Override
-    public ListMultimap<AbstractProject, Future<AbstractBuild>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-        ListMultimap<AbstractProject, Future<AbstractBuild>> futures = super.perform2(build, launcher, listener);
+    public ListMultimap<Job, Future<Run>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+        ListMultimap<Job, Future<Run>> futures = super.perform2(build, launcher, listener);
         if(block==null) return ArrayListMultimap.create();
         return futures;
     }
 
     @Override
-    protected Future schedule(AbstractBuild<?, ?> build, AbstractProject project, List<Action> list) throws InterruptedException, IOException {
+    protected Future schedule(AbstractBuild<?, ?> build, Job project, List<Action> list) throws InterruptedException, IOException {
         if (block!=null) {
             while (true) {
                 // add DifferentiatingAction to make sure this doesn't get merged with something else,
@@ -70,7 +65,8 @@ public class BlockableBuildTriggerConfig extends BuildTriggerConfig {
                 // if we fail to add the item to the queue, wait and retry.
                 // it also means we have to force quiet period = 0, or else it'll never leave the queue
                 Future f = schedule(build, project, 0, list);
-                //when a project is disabled or the configuration is not yet saved f will always be null and we'ure caught in a loop, therefore we need to check for it
+
+                //when a project is disabled or the configuration is not yet saved f will always be null and we're caught in a loop, therefore we need to check for it
                 if (f!=null || (f==null && !project.isBuildable())){
                     return f;
                 }
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
@@ -55,8 +55,15 @@ public class BlockableBuildTriggerConfig extends BuildTriggerConfig {
     }
 
     @Override
-    public ListMultimap<Job, Future<Run>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-        ListMultimap<Job, Future<Run>> futures = super.perform2(build, launcher, listener);
+    public ListMultimap<AbstractProject, Future<AbstractBuild>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+        ListMultimap<AbstractProject, Future<AbstractBuild>> futures = super.perform2(build, launcher, listener);
+        if(block==null) return ArrayListMultimap.create();
+        return futures;
+    }
+
+    @Override
+    public ListMultimap<Job, Future<Run>> perform3(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+        ListMultimap<Job, Future<Run>> futures = super.perform3(build, launcher, listener);
         if(block==null) return ArrayListMultimap.create();
         return futures;
     }
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
@@ -3,8 +3,15 @@ package hudson.plugins.parameterizedtrigger;
 import com.google.common.collect.ImmutableList;
 import hudson.Extension;
 import hudson.Launcher;
-import hudson.model.*;
+import hudson.model.AbstractBuild;
+import hudson.model.AbstractProject;
+import hudson.model.Action;
+import hudson.model.BuildListener;
 import hudson.model.Cause.UpstreamCause;
+import hudson.model.Hudson;
+import hudson.model.Job;
+import hudson.model.Node;
+import hudson.model.Run;
 import jenkins.model.ParameterizedJobMixIn;
 import org.kohsuke.stapler.DataBoundConstructor;
 
@@ -65,8 +72,7 @@ public class BlockableBuildTriggerConfig extends BuildTriggerConfig {
                 // if we fail to add the item to the queue, wait and retry.
                 // it also means we have to force quiet period = 0, or else it'll never leave the queue
                 Future f = schedule(build, project, 0, list);
-
-                //when a project is disabled or the configuration is not yet saved f will always be null and we're caught in a loop, therefore we need to check for it
+                // When a project is disabled or the configuration is not yet saved f will always be null and we're caught in a loop, therefore we need to check for it
                 if (f!=null || (f==null && !project.isBuildable())){
                     return f;
                 }
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
@@ -4,9 +4,14 @@ import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Launcher;
 import hudson.Util;
-import hudson.model.*;
+import hudson.model.AbstractBuild;
+import hudson.model.AbstractProject;
+import hudson.model.BuildListener;
+import hudson.model.DependecyDeclarer;
+import hudson.model.DependencyGraph;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
+import hudson.model.Job;
 import hudson.tasks.Notifier;
 import hudson.tasks.Publisher;
 import org.kohsuke.stapler.DataBoundConstructor;
@@ -58,6 +63,7 @@ public class BuildTrigger extends Notifier implements DependecyDeclarer {
 			for (Job j : jobs) {
 				if (!(j instanceof AbstractProject)) {
 					hasNonAbstractProject = true;
+					break;
 				}
 			}
 			// Fire this config's projects if not already fired
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
@@ -50,28 +50,27 @@ public class BuildTrigger extends Notifier implements DependecyDeclarer {
 	}
 
 	@Override @SuppressWarnings("deprecation")
-	public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
-			BuildListener listener) throws InterruptedException, IOException {
+	public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
 
-		HashSet<BuildTriggerConfig> alreadyFired = new HashSet<BuildTriggerConfig>();
+        HashSet<BuildTriggerConfig> alreadyFired = new HashSet<BuildTriggerConfig>();
 
-		// If this project has non-abstract projects, we need to fire them
-		for (BuildTriggerConfig config : configs) {
-			boolean hasNonAbstractProject = false;
+        // If this project has non-abstract projects, we need to fire them
+        for (BuildTriggerConfig config : configs) {
+            boolean hasNonAbstractProject = false;
 
-			List<Job> jobs = config.getProjectList(build.getRootBuild().getProject().getParent(), build.getEnvironment(listener));
-			for (Job j : jobs) {
-				if (!(j instanceof AbstractProject)) {
-					hasNonAbstractProject = true;
-					break;
-				}
-			}
-			// Fire this config's projects if not already fired
-			if (hasNonAbstractProject) {
-				config.perform(build, launcher, listener);
-				alreadyFired.add(config);
-			}
-		}
+            List<Job> jobs = config.getProjectList(build.getRootBuild().getProject().getParent(), build.getEnvironment(listener));
+            for (Job j : jobs) {
+                if (!(j instanceof AbstractProject)) {
+                    hasNonAbstractProject = true;
+                    break;
+                }
+            }
+            // Fire this config's projects if not already fired
+            if (hasNonAbstractProject) {
+                config.perform(build, launcher, listener);
+                alreadyFired.add(config);
+            }
+        }
 
         if (canDeclare(build.getProject())) {
             // job will get triggered by dependency graph, so we have to capture buildEnvironment NOW before
@@ -80,9 +79,9 @@ public class BuildTrigger extends Notifier implements DependecyDeclarer {
             build.addAction(new CapturedEnvironmentAction(env));
         } else {  // Not using dependency graph
             for (BuildTriggerConfig config : configs) {
-				if (!alreadyFired.contains(config)) {
-					config.perform(build, launcher, listener);
-				}
+                if (!alreadyFired.contains(config)) {
+                     config.perform(build, launcher, listener);
+                }
             }
         }
 
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
@@ -58,7 +58,7 @@ public class BuildTrigger extends Notifier implements DependecyDeclarer {
         for (BuildTriggerConfig config : configs) {
             boolean hasNonAbstractProject = false;
 
-            List<Job> jobs = config.getProjectListGeneral(build.getRootBuild().getProject().getParent(), build.getEnvironment(listener));
+            List<Job> jobs = config.getJobs(build.getRootBuild().getProject().getParent(), build.getEnvironment(listener));
             for (Job j : jobs) {
                 if (!(j instanceof AbstractProject)) {
                     hasNonAbstractProject = true;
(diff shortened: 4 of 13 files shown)


src/main/java/hudson/plugins/parameterizedtrigger/AbstractBuildParameterFactory.java
```diff
@@ -1,9 +1,7 @@
 package hudson.plugins.parameterizedtrigger;
 
 import hudson.ExtensionPoint;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractDescribableImpl;
-import hudson.model.TaskListener;
+import hudson.model.*;
```
jtnord: prefer explicit changes over * imports.
This file has no (meaningful) changes.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -50,18 +47,39 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+				if (!(j instanceof AbstractProject)) {
+					hasNonAbstractProject = true;
+				}
+			}
+			// Fire this config's projects if not already fired
+			if (hasNonAbstractProject) {
+				config.perform(build, launcher, listener);
+				alreadyFired.add(config);
+			}
+		}
```
jtnord: the indentation here is completely out of sync with the rest of the method.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -50,18 +47,39 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+		HashSet<BuildTriggerConfig> alreadyFired = new HashSet<BuildTriggerConfig>();
+
+		// If this project has non-abstract projects, we need to fire them
+		for (BuildTriggerConfig config : configs) {
+			boolean hasNonAbstractProject = false;
+
+			List<Job> jobs = config.getProjectList(build.getRootBuild().getProject().getParent(), build.getEnvironment(listener));
+			for (Job j : jobs) {
+				if (!(j instanceof AbstractProject)) {
+					hasNonAbstractProject = true;
```
jtnord: seems like you should break here.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -50,18 +47,39 @@ public BuildStepMonitor getRequiredMonitorService() {
...
         if (canDeclare(build.getProject())) {
             // job will get triggered by dependency graph, so we have to capture buildEnvironment NOW before
             // hudson.model.AbstractBuild.AbstractBuildExecution#cleanUp is called and reset
             EnvVars env = build.getEnvironment(listener);
             build.addAction(new CapturedEnvironmentAction(env));
-        } else {
+        } else {  // Not using dependency graph
             for (BuildTriggerConfig config : configs) {
-                config.perform(build, launcher, listener);
+				if (!alreadyFired.contains(config)) {
```
jtnord: indentation...

jtnord: maybe I missed something - but there is no point in this check  - or the loop above as you will always call config.perform with the same arguments - either here (if its not `alreadyFired`, or above.)

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -307,9 +293,8 @@ private static void resolveProject(AbstractBuild build, SubProjectData subProjec
                     List<Action> actions = getBaseActions(
                             ImmutableList.<AbstractBuildParameters>builder().addAll(configs).addAll(addConfigs).build(),
                             build, listener);
-                    for (AbstractProject project : getProjectList(build.getRootBuild().getProject().getParent(),env)) {
+					for (Job project : getProjectList(build.getRootBuild().getProject().getParent(),env)) {
```
jtnord: indentation

jtnord:
:bug:s for the import *, and not matching the current indentation scheme.

Theres also an open question on the `BuildTrigger.java`

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -70,9 +88,11 @@ public void buildDependencyGraph(AbstractProject owner, DependencyGraph graph) {
 		// Can only add dependencies in Hudson 1.341 or higher
 		if (!canDeclare(owner)) return;
 
-		for (BuildTriggerConfig config : configs)
-			for (AbstractProject project : config.getProjectList(owner.getParent(),null))
+		for (BuildTriggerConfig config : configs) {
+			List<AbstractProject> projectList = Util.filter(config.getProjectList(owner.getParent(), null), AbstractProject.class);
```
amuniz: Are you expecting to get Workflow jobs here? If so, I think it will not work.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
svanoort: @jtnord It's subtle -- the canDeclare method is sufficient for most cases, where all builds will be handled via the dependency graph and all projects extend AbstractProject.  

We're only explicitly calling the trigger perform method if the trigger includes workflow jobs (which DependencyGraph does not support) OR  if canDeclare is false... in cases where both are true, I'm trying to avoid performing the same build twice.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
svanoort: @amuniz Thanks for the feedback.  Workflow jobs are explicitly fired via the BuildTrigger.perform method; they don't participate in the DependencyGraph mechanism (core APIs don't support WorkflowJobs -- it's been raised as a potential enhancement, but not yet).

src/main/java/hudson/plugins/parameterizedtrigger/AbstractBuildParameterFactory.java
```diff
@@ -24,10 +22,10 @@
      *      The build which the parameterized trigger is configured and executing.
      * @param listener
      *      Connected to the build output.
-     *      
+     *
      * @return can be empty but never null.
      */
-    public abstract List<AbstractBuildParameters> getParameters(AbstractBuild<?,?> build, TaskListener listener)
+    public abstract List<AbstractBuildParameters> getParameters(AbstractBuild<?, ?> build, TaskListener listener)
```
oleg-nenashev: Please avoid such formatting changes if you don't change the line. They make the review and merges more difficult

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -3,14 +3,9 @@
 import com.google.common.collect.ImmutableList;
 import hudson.Extension;
 import hudson.Launcher;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractProject;
-import hudson.model.Action;
-import hudson.model.BuildListener;
+import hudson.model.*;
```
oleg-nenashev: No, the common practice in Jenkins is to refer all imports to prevent conflicts of new classes

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -53,14 +48,14 @@ public BlockingBehaviour getBlock() {
     }
 
     @Override
-    public ListMultimap<AbstractProject, Future<AbstractBuild>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-        ListMultimap<AbstractProject, Future<AbstractBuild>> futures = super.perform2(build, launcher, listener);
+    public ListMultimap<Job, Future<Run>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
```
oleg-nenashev: Binary backward compatibility breach. Collections cannot be promoted in such way

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -111,7 +97,7 @@ public boolean getTriggerWithNoParameters() {
      * @deprecated
      *      Use {@link #getProjectList(ItemGroup, EnvVars)}
      */
-    public List<AbstractProject> getProjectList(EnvVars env) {
+    public List<Job> getProjectList(EnvVars env) {
```
oleg-nenashev: Backward compatibility

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -120,9 +106,9 @@ public boolean getTriggerWithNoParameters() {
      * @param context
      *      The container with which to resolve relative project names.
      */
-	public List<AbstractProject> getProjectList(ItemGroup context, EnvVars env) {
-        List<AbstractProject> projectList = new ArrayList<AbstractProject>();
-        projectList.addAll(Items.fromNameList(context, getProjects(env), AbstractProject.class));
+	public List<Job> getProjectList(ItemGroup context, EnvVars env) {
```
oleg-nenashev: Backward compatibility

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -322,19 +307,18 @@ private static void resolveProject(AbstractBuild build, SubProjectData subProjec
         return Collections.emptyList();
 	}
 
-    public ListMultimap<AbstractProject, Future<AbstractBuild>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+    public ListMultimap<Job, Future<Run>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
```
oleg-nenashev: Backward compatibility

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -403,15 +387,35 @@ protected Cause createUpstreamCause(AbstractBuild<?, ?> build) {
         return new UpstreamCause((Run) build);
```
oleg-nenashev: You should remove obsolete class casts

src/main/java/hudson/plugins/parameterizedtrigger/DefaultParameterValuesActionsTransform.java
```diff
@@ -1,11 +1,6 @@
 package hudson.plugins.parameterizedtrigger;
 
-import hudson.model.AbstractProject;
-import hudson.model.Action;
-import hudson.model.ParametersAction;
-import hudson.model.ParameterDefinition;
-import hudson.model.ParameterValue;
-import hudson.model.ParametersDefinitionProperty;
+import hudson.model.*;
```
oleg-nenashev: Nope, it should not be done in such way

oleg-nenashev:
:bug: for binary compatibility changes. If you want to go in such way, you'll need much more justification of the change.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
jtnord: so `canDeclare` will always be false for workflow (as it doesn't support the dependencyGraph), so the `else` will always be entered.  So I'm still confused.

svanoort: @jtnord Fair enough, it's rather confusing!  

The upstream project is always an AbstractProject (you can't add the trigger on a workflow job, but workflow has its own mechanism to fire downstream builds).  canDeclare is invoked on the *upstream* project, an AbstractProject, simply to see if the *upstream* can participate in the DependencyGraph, which might be true or false.  

Regardless of this, we need to ensure that workflow jobs are triggered.    I'm not going to guarantee there aren't edge cases until the tests are reviewed, but this part at least should work.

I hope that makes the rationale a bit clearer?  I'd be happy to do a hangout if it still seems iffy to you.

svanoort:
@oleg-nenashev Can you please explain a bit more about where we expect to hit issues from changing method signatures and breaking binary compatibility?

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -403,15 +405,35 @@ protected Cause createUpstreamCause(AbstractBuild<?, ?> build) {
...
+
+        // Includes both traditional projects via AbstractProject and Workflow Job
+        if (project instanceof ParameterizedJobMixIn.ParameterizedJob) {
+            final ParameterizedJobMixIn<?, ?> parameterizedJobMixIn = new ParameterizedJobMixIn() {
+                @Override
+                protected Job<?, ?> asJob() {
+                    return project;
+                }
+            };
+            return parameterizedJobMixIn.scheduleBuild2(quietPeriod, queueActions.toArray(new Action[queueActions.size()]));
```
jglick: FYI https://github.com/jenkinsci/jenkins/pull/1771 would be nicer here. I would recommend leaving in a `TODO` comment to that effect.

src/main/java/hudson/plugins/parameterizedtrigger/DefaultParameterValuesActionsTransform.java
jglick: There is no general styleguide for Jenkins plugins. My personal preference is to avoid star imports (except for `import static`), and keep all imports simply sorted in one block—after all, imports are for `javac` and your IDE to read, not for humans.

pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
+        <version>1.607</version>
```
amuniz: IMO Better to depend on the next (nearest) LTS, which is 1.609

jglick: 1.580.1 was an LTS so this is fine. If there is no particular reason to require a newer LTS line we generally do not do so.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -109,24 +114,36 @@ public boolean getTriggerWithNoParameters() {
...
+        return getProjectList(null, env);
     }
 
     /**
+     * Deprecated: get all projects that are AbstractProject instances
      * @param env Environment variables from which to expand project names; Might be {@code null}.
      * @param context
      *      The container with which to resolve relative project names.
      */
+    @Deprecated // Prefer getProjectListGeneral since it can return implementations of the more general Job class
```
amuniz: I would add ```@deprecated``` javadoc also (more visible from IDEs when developing and Javadoc pages itself)

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -322,17 +339,35 @@ private static void resolveProject(AbstractBuild build, SubProjectData subProjec
         return Collections.emptyList();
 	}
 
+    @Deprecated
```
amuniz: Same here (```@deprecated``` javadoc)

pom.xml
svanoort: 1.607 was the earliest I could find that supported  jenkins.model.queue.AsynchronousExecution -- I've gone ahead and bumped to 1.609 (and re-run tests) so it's tied to an LTS.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
svanoort: @amuniz Thanks, good point -- added this in latest commit

amuniz:
:bee:

pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
+        <version>1.609</version>
```
jglick: As I responded to @amuniz this does not seem to be necessary and so should probably be reverted (unless you can offer an explanation for why you need a newer baseline).

jglick:
:bug: for the apparently gratuitous baseline update.

pom.xml
amuniz: @jglick I misunderstood your message, I thought you agreed with the change. Anyway I think is a good practice to use LTS baselines since you are relying on a known stable version to develop the plugin.

svanoort:
@jglick It's not gratuitous -- as I explained in an earlier comment reply, there is an API not present in 1.580.1, specifically jenkins.model.queue.AsynchronousExecution.  1.607 was the earliest I could find that supported  this, and 1.609 is next LTS after that.

jglick:
> there is an API not present in 1.580.1

Sorry, I missed that, since it was in a comment in an outdated diff, and the commit changing the baseline did not use any new API. It is a good practice to mention the reason for the baseline in a comment in the line specifying the baseline. Where are you using `AsynchronousExecution`?

pom.xml
```diff
@@ -63,6 +63,27 @@
           <version>1.3</version>
           <scope>test</scope>
         </dependency>
+
+        <!-- Workflow plugin imports are used to test compatibility -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-job</artifactId>
+            <version>1.8</version>
```
jglick: Nicer to introduce a POM property for the version.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -109,24 +114,38 @@ public boolean getTriggerWithNoParameters() {
...
+        return Util.filter(getProjectListGeneral(context,env), AbstractProject.class);
 	}
 
     /**
+     * Get list of all projects, including workflow job types
+     * @param env Environment variables from which to expand project names; Might be {@code null}.
+     * @param context
+     *      The container with which to resolve relative project names.
+     */
+    public List<Job> getProjectListGeneral(ItemGroup context, EnvVars env) {
```
jglick: `getJobList` might be a more natural name.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -400,18 +441,40 @@ protected Cause createUpstreamCause(AbstractBuild<?, ?> build) {
                 return new UpstreamCause((Run<?,?>)promotion.getTarget());
             }
         }
-        return new UpstreamCause((Run) build);
+        return new UpstreamCause(build);
     }
 
-    protected Future schedule(AbstractBuild<?, ?> build, AbstractProject project, int quietPeriod, List<Action> list) throws InterruptedException, IOException {
+    protected Future schedule(AbstractBuild<?, ?> build, final Job project, int quietPeriod, List<Action> list) throws InterruptedException, IOException {
+        // TODO Once it's in core and LTS is out, switch to use new ParameterizedJobMixIn convenience method
```
jglick: Recommend specifying the since-version, `1.621`, in the comment so it is quick for a future plugin maintainer to tell whether they should use it.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -400,18 +441,40 @@ protected Cause createUpstreamCause(AbstractBuild<?, ?> build) {
                 return new UpstreamCause((Run<?,?>)promotion.getTarget());
             }
         }
-        return new UpstreamCause((Run) build);
+        return new UpstreamCause(build);
     }
 
-    protected Future schedule(AbstractBuild<?, ?> build, AbstractProject project, int quietPeriod, List<Action> list) throws InterruptedException, IOException {
+    protected Future schedule(AbstractBuild<?, ?> build, final Job project, int quietPeriod, List<Action> list) throws InterruptedException, IOException {
```
jglick: `@CheckForNull`

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +132,65 @@ public void testGetProjectListStatic() throws Exception {
 
     }
 
+    public void testGetProjectListWithWorkflow() throws Exception {
+        Project<?, ?> masterProject = createFreeStyleProject("project");
+        WorkflowJob p = (WorkflowJob) Jenkins.getInstance().createProject(WorkflowJob.class, "workflowproject");
```
jglick: Redundant cast. Also remember that from a `HudsonTestCase` you can use simply `jenkins` rather than `Jenkins.getInstance()`.

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +132,65 @@ public void testGetProjectListStatic() throws Exception {
...
+        List<AbstractProject> projects = masterConfig.getProjectList(masterProject.getParent(), null);
+        assertEquals(1, projects.size());
+        assertFalse("Projects should NOT include workflow job", projects.contains(p));
+        assertTrue("Projects should include non-workflow job", projects.contains(subProject1));
+    }
+
+    public void testBuildWithWorkflowProjects() throws Exception {
+        Project<?, ?> masterProject = createFreeStyleProject("project");
+        WorkflowJob workflowProject = (WorkflowJob) Jenkins.getInstance().createProject(WorkflowJob.class, "workflowproject");
+        workflowProject.setDefinition(new CpsFlowDefinition("node { echo myParam; }"));
```
jglick: `node` block unnecessary BTW.

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +132,65 @@ public void testGetProjectListStatic() throws Exception {
...
+
+        // Trigger a normal and workflow project
+        String projectToTrigger = "subproject1, workflowproject";
+        List<AbstractBuildParameters> buildParameters = new ArrayList<AbstractBuildParameters>();
+        buildParameters.add(new CurrentBuildParameters());
```
jglick: Why?

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +132,65 @@ public void testGetProjectListStatic() throws Exception {
...
+
+        BlockingBehaviour neverFail = new BlockingBehaviour("never", "never", "never");
+        BlockableBuildTriggerConfig masterConfig = new BlockableBuildTriggerConfig(projectToTrigger, neverFail, buildParameters);
+
+        addParameterizedTrigger(masterProject, masterConfig);
+
+        Project subProject1 = createFreeStyleProject("subproject1");
+        subProject1.setQuietPeriod(0);
+
+        masterProject.scheduleBuild2(0, new Cause.UserCause()).get();
```
jglick: You can omit the `Cause` as a rule.

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +132,65 @@ public void testGetProjectListStatic() throws Exception {
...
+
+        // Check all builds triggered correctly
+        assertEquals(1, workflowProject.getBuilds().size());
+        assertEquals(1, subProject1.getBuilds().size());
+
+        // Verify workflow job completed successfully and that it was able to use the parameter set in trigger
+        WorkflowRun workflowRun = workflowProject.getBuilds().get(0);
+        assertEquals(Result.SUCCESS.ordinal, workflowRun.getResult().ordinal);
```
jglick:
```java
assertBuildStatusSuccess(workflowRun);
```

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +132,65 @@ public void testGetProjectListStatic() throws Exception {
...
+        assertTrue("Parameter was not passed correctly to workflow job!", workflowRun.getLog(100).contains("GOOBER"));
```
jglick:
```java
assertLogContains("GOOBER", workflowRun);
```

src/test/java/hudson/plugins/parameterizedtrigger/test/TriggerBuilderTest.java
```diff
@@ -211,6 +220,36 @@ public void testBlockingTriggerWithDisabledProjects() throws Exception {
...
+        p.setDefinition(new CpsFlowDefinition("println('hello')"));
+
+        Project<?, ?> triggerProject = createFreeStyleProject("projectA");
+        TriggerBuilder triggerBuilder = new TriggerBuilder(createTriggerConfig("project1"));
+        triggerProject.getBuildersList().add(triggerBuilder);
+        triggerProject.scheduleBuild2(0).get();
+
+        assertNotNull(triggerProject.getLastBuild());
+        assertEquals(1, p.getBuilds().size());
+        assertNotNull(p.getLastBuild());
```
jglick: Does this actually test anything different from `BuildTriggerConfigTest.testBuildWithWorkflowProjects`?

jglick:
:bug: for the (probably) unnecessary baseline update. All else are minor comments.

pom.xml
stephenc: Makes sense as the version is repeated 3 times, as long as you are releasing all three of those plugins at the same time. If/when you refactor them into separate git repositories with independent release lifecycles then it would make sense to list explicit versions... but for now a property will ensure consistent versioning

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
svanoort: Modified, change will be visible when latest commit pushed

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
svanoort: Why what?

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
svanoort: Done, will be in latest commit when pushed.

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
svanoort: Thanks for the note -- will be in latest commit when pushed.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
svanoort: Done, will be in latest commit when pushed

pom.xml
svanoort: Thanks for feedback -- good point, will be in latest commit once pushed

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
jglick: I am wondering why you are using `CurrentBuildParameters` at all, when you are passing no parameters to `masterProject.scheduleBuild2`. The test would seem to stand on just `PredefinedBuildParameters`.

jglick:
:bee: other than the minor outstanding comments.

pom.xml
```diff
@@ -160,6 +181,7 @@
     <properties>
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <workflow.version>1.4</workflow.version>
```
stephenc: :ant: whitespace

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
stephenc: agree... the method name is not one I would like to bed into this api, `getJobList` is much better

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -48,20 +50,41 @@ public BuildStepMonitor getRequiredMonitorService() {
 	}
 
 	@Override @SuppressWarnings("deprecation")
-	public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
-			BuildListener listener) throws InterruptedException, IOException {
+	public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+
+        HashSet<BuildTriggerConfig> alreadyFired = new HashSet<BuildTriggerConfig>();
```
oleg-nenashev: Minor: wrong code indenting

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -70,9 +93,11 @@ public void buildDependencyGraph(AbstractProject owner, DependencyGraph graph) {
 		// Can only add dependencies in Hudson 1.341 or higher
 		if (!canDeclare(owner)) return;
 
-		for (BuildTriggerConfig config : configs)
-			for (AbstractProject project : config.getProjectList(owner.getParent(),null))
+		for (BuildTriggerConfig config : configs) {
+			List<AbstractProject> projectList = config.getProjectList(owner.getParent(), null);
+			for (AbstractProject project : projectList)
```
oleg-nenashev: please use brackets even for single-line cycles

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
oleg-nenashev: +1

stephenc:
:bug: for the `getProjectListGeneral` method name. It's littering a API that other plugins may end up using where `getJobList` would be more appropriate

other than that it's all :ant:s that can be ignored

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -109,24 +115,38 @@ public boolean getTriggerWithNoParameters() {
...
 
     /**
+     * Get list of all projects, including workflow job types
+     * @param env Environment variables from which to expand project names; Might be {@code null}.
+     * @param context
+     *      The container with which to resolve relative project names.
+     */
+    public List<Job> getProjectListGeneral(ItemGroup context, EnvVars env) {
+        List<Job> projectList = new ArrayList<Job>();
+        projectList.addAll(Items.fromNameList(context, getProjects(env), Job.class));
```
oleg-nenashev: Seems <code>Items.fromNameList</code> does not care about full names => I suspect it won't work for folders

oleg-nenashev:
:bug: for the folders integration issue (please correct me if I'm wrong)

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
svanoort: Are you sure?  I ask because the Items.fromNameList logic was present before my commits (I did not change that) and as far as I can tell, dates back to at least 2012.

oleg-nenashev: I've investigated deeper and discovered that <code>Jenkins.getItem</code> now falls back to <code>Jenkins.getItemByFullName()</code> starting from 1.460. This implementation should work properly even there's a slight performance improvement potential

oleg-nenashev:
removed the bug after the discussion

jglick:
`BuildTriggerConfigTest` uncompilable.

jglick:
:bug: for failure to include `src/test/java/` in rename refactoring :open_mouth:

stephenc:
:bug: build failing

stephenc:
:bee: (conditional on Jenkins saying it builds)

jglick:
:bee: (also conditional on tests passing of course)

src/main/java/hudson/plugins/parameterizedtrigger/DefaultParameterValuesActionsTransform.java
jtnord: if I was a maintainer of this I would reject such a change.
You may get lucky...

src/main/java/hudson/plugins/parameterizedtrigger/ProjectSpecificParameterValuesActionTransform.java
```diff
@@ -1,17 +1,6 @@
...
-import hudson.model.Action;
-import hudson.model.ParametersAction;
-import hudson.model.ParameterDefinition;
-import hudson.model.ParametersDefinitionProperty;
-import hudson.model.Queue;
-import hudson.model.ParameterValue;
-import hudson.model.SimpleParameterDefinition;
-import hudson.model.StringParameterDefinition;
-import hudson.model.StringParameterValue;
+import hudson.model.*;
```
jtnord: if I was the maintainer of this plugin I would not be happy with this change.
YMMV.

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
```diff
@@ -121,6 +126,64 @@ public void testGetProjectListStatic() throws Exception {
 
     }
 
+    public void testGetProjectListWithWorkflow() throws Exception {
+        Project<?, ?> masterProject = createFreeStyleProject("project");
+        WorkflowJob p = jenkins.createProject(WorkflowJob.class, "workflowproject");
+        p.setDefinition(new CpsFlowDefinition("println('hello')"));
```
jtnord: nit: should this not be echo - so it uses workflow rather than just some groovy?

jtnord:
:bee: on the code.  If I was the maintainer I would not be happy with the change to `import *` - however I am not and you may get lucky :smile:

svanoort:
@jtnord I fixed the imports -- looks like one slipped through from earlier.   
I've since told my IDE to stop trying to be "helpful" by collapsing imports unless it has 99 from the same pkg.  

@jglick This includes the fix for the UI checker irritation

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerConfigTest.java
jglick: Actually Workflow hacks things a bit so that `println` is just an alias for `echo`. It is preferable to use `echo` directly.

jglick:
:bee:

jglick:
:bee:

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -542,17 +609,28 @@ public FormValidation doCheckProjects(@AncestorInPath AbstractProject<?,?> proje
                 if (StringUtils.isNotBlank(projectName)) {
                 	Item item = Jenkins.getInstance().getItem(projectName,project,Item.class); // only works after version 1.410
                     if(item==null){
-                        return FormValidation.error(Messages.BuildTrigger_NoSuchProject(projectName,AbstractProject.findNearest(projectName).getName()));
+                        Item nearest = Items.findNearest(Job.class, projectName, Jenkins.getInstance());
+                        String alternative = nearest != null ? nearest.getRelativeNameFrom(project) : "?";
+                        return FormValidation.error(Messages.BuildTrigger_NoSuchProject(projectName, alternative));
                     }
-                    if(!(item instanceof AbstractProject)){
+                    if(!(item instanceof Job) || !(item instanceof ParameterizedJobMixIn.ParameterizedJob)) {
```
amuniz: What's intended to be handled here? Would it be possible that the item were not a Job/AbstractProject?

amuniz:
I think this is :bee: (beyond the little question I put there)

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
jglick: Sure, for example a `Folder` is an `Item` but not a `Job`, and an [`ExternalJob`](https://github.com/jenkinsci/external-monitor-job-plugin/blob/c1725b597003f8ce1701e6e2fe83d6dd2f424671/src/main/java/hudson/model/ExternalJob.java#L45) is a `Job` but not a `ParameterizedJob`.

amuniz: OK. Thanks.

jglick:
@reviewbybees done

(since @oleg-nenashev forgot to add an amending bee before going on vacation)

svanoort:
After checking with active community members to see if they have concerns, I'm going ahead and merging this since we have users waiting on it.

oleg-nenashev:
@svanoort 
According to the Jenkins process, you should try to contact the plugin's owner and wait for 1 week for the response before going forward. IIRC @ikedam was the de-facto owner of the plugin for a while.

Anyway, waiting for a response in the mailing list on the weekend was not enough IMHO.

svanoort:
@oleg-nenashev  To capture not visible discussion, I'd reached out to KK (who is still formally listed as the maintainer, even if community members) and olivergondza (who's done a lot of the releases and commits recently, more than ikedam) with positive responses from both.  In this case we also have at least one larger Jenkins user waiting on this to become available, so I'm trying to juggle competing concerns here.  

I'll certainly bear in mind that community process is to wait a week (in absence of actively solicited feedback as I received in this case), and will try to make that feedback more visible in general.

oleg-nenashev:
Thanks for the clarification and for all your efforts to get it merged!

PR #90, opened 2015-08-21
Workflow build fails when futures is empty.

tonyfinn:
The type of projectList was changed to `List<Job>` in #87, but items from the iterator were still being cast to AbstractProject for the log message.


jglick:
:+1: and CC @svanoort

svanoort:
Hi, this looks good but lacks tests.  I went ahead and added tests (and verified manually), adding this to https://github.com/jenkinsci/parameterized-trigger-plugin/pull/92

Would suggest we continue there

tonyfinn:
Ok, I'll close this PR then.

PR #91, opened 2015-08-27
Miscellaneous minor fixes and improvements

sschuberth:
Mainly fix unit tests for default encoding on Windows plus some other minor fixes and improvements.

### DIFF
--- a/src/main/resources/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig/config.jelly
+++ b/src/main/resources/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig/config.jelly
@@ -25,13 +25,7 @@ THE SOFTWARE.
   <f:entry title="${%Projects to build}" field="projects">
     <f:textbox />
   </f:entry>
-  <!-- TODO: replace with <f:optionalProperty> -->
-  <f:optionalBlock field="block" title="${%Block until the triggered projects finish their builds}" checked="${instance.block!=null}">
-    <j:set var="descriptor" value="${app.getDescriptorOrDie(descriptor.getPropertyType(field).clazz)}" />
-    <j:set var="instance" value="${instance[field]}"/>
-    <st:include from="${descriptor}" page="${descriptor.configPage}" />
-  </f:optionalBlock>
-
+  <f:optionalProperty field="block" title="${%Block until the triggered projects finish their builds}" />
   <f:block>
     <f:hetero-list name="configs" hasHeader="true"
                    descriptors="${descriptor.getBuilderConfigDescriptors()}"
--- a/src/test/java/hudson/plugins/parameterizedtrigger/test/FileBuildParameterFactoryTest.java
+++ b/src/test/java/hudson/plugins/parameterizedtrigger/test/FileBuildParameterFactoryTest.java
@@ -96,7 +96,7 @@ public class FileBuildParameterFactoryTest extends HudsonTestCase {
             }
         });
 
-        // add Trigger builder, with file paramter factory
+        // add Trigger builder, with file parameter factory
         projectA.getBuildersList().add(createTriggerBuilder(projectB, NoFilesFoundEnum.SKIP));
 
         projectA.scheduleBuild2(0).get();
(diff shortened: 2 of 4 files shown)


sschuberth:
@svanoort Would you mind having a look?

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -30,7 +30,6 @@
  */
 public class BlockableBuildTriggerConfig extends BuildTriggerConfig {
     private final BlockingBehaviour block;
-    public boolean buildAllNodesWithLabel;
```
oleg-nenashev: Breaks binary compatibility. If unused, should be marked as deprecated

src/main/resources/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig/config.jelly
```diff
@@ -25,13 +25,7 @@ THE SOFTWARE.
...
     <f:textbox />
   </f:entry>
-  <!-- TODO: replace with <f:optionalProperty> -->
-  <f:optionalBlock field="block" title="${%Block until the triggered projects finish their builds}" checked="${instance.block!=null}">
-    <j:set var="descriptor" value="${app.getDescriptorOrDie(descriptor.getPropertyType(field).clazz)}" />
-    <j:set var="instance" value="${instance[field]}"/>
-    <st:include from="${descriptor}" page="${descriptor.configPage}" />
-  </f:optionalBlock>
-
+  <f:optionalProperty field="block" title="${%Block until the triggered projects finish their builds}" />
```
oleg-nenashev: Did you manually test it?

src/test/resources/hudson/plugins/parameterizedtrigger/test/BuildInfoExporterTest/testMigrateFrom221/jobs/upstream/config.xml
```diff
@@ -104,11 +102,10 @@
               <color>RED</color>
             </failureThreshold>
           </block>
-          <buildAllNodesWithLabel>false</buildAllNodesWithLabel>
```
oleg-nenashev: In this test change you hide the problem you introduce by breaking the compatibility

oleg-nenashev:
:-1: for merge in the current state

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
sschuberth: I'll drop this change as I not not want to mix this PR of mine with code deprecation.

src/main/resources/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig/config.jelly
sschuberth: Yes, it works according to my test.

src/test/resources/hudson/plugins/parameterizedtrigger/test/BuildInfoExporterTest/testMigrateFrom221/jobs/upstream/config.xml
sschuberth: Dropping this change, too.

sschuberth:
@oleg-nenashev I've updated the PR to address your comments.

svanoort:
Apologies for the delay.  It looks like my filter that was supposed to tag direct github mentions (so they don't get buried in the 50+ per day) needs some tweaks, and this slipped through. 

:+1:  from me for this, and I'll merge tomorrow if no objections are raised

sschuberth:
I've pushed an update that removes the multi-byte assertions for non-multi-byte default character sets, as I believe there's no point in comparing against "???"-like strings.

oleg-nenashev:
:+1:

sschuberth:
@svanoort I think it's ready now <wink> ;-)

svanoort:
:+1:  and merge as a maintainer

PR #92, opened 2015-09-01
[JENKINS-30040] Logging exception running triggering nonblocking builds

svanoort:
Builds on top of fix in [PR #90](https://github.com/jenkinsci/parameterized-trigger-plugin/pull/90) but includes tests.  I've manually verified this works correctly from the UI as well. 

@tonyfinn @reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
@@ -39,6 +39,7 @@ import hudson.tasks.Builder;
 import hudson.model.Job;
 import hudson.model.Run;
 import hudson.util.IOException2;
+import org.kohsuke.accmod.Restricted;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -157,10 +158,12 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
         return buildStepResult;
     }
 
-    private String getProjectListAsString(List<Job> projectList){
+    // Public but restricted so we can add tests without completely changing the tests package
+    @Restricted(value=org.kohsuke.accmod.restrictions.NoExternalUse.class)
```
oleg-nenashev: NIT: would be better to just import the class, value= is also not required

oleg-nenashev:
:bee:

tfennelly:
:bee:

svanoort:
@reviewbybees done.

svanoort:
@ikedam Any thoughts on this, as maintainer?

ikedam:
@svanoort 
I no longer maintain this plugin.
Please contact the current maintainer.

svanoort:
Thanks!  

@olivergondza Would you have any issues if this were merged / released?  It's a fairly trivial fix for an irritating bug.

olivergondza:
Releasing 2.29 right now.

PR #93, opened 2015-09-25
Append build description to each build in the Subprojects view

dbarbieri:
Since the Parameter factory runs the same job with different parameters, it is hard to discern build which is which in the Subproject view when only a job name and build number are shown. It would be more informative if we include the build description too. You can use the Description Setter Plugin to auto set the build description to any unique parameters passed in from the Parameter Factory, allowing the end user to choose how this page should be displayed.

1 caviot: It seems the text() function turns strings from "<a href" to "&lt;a href". I would like it if any html links in the build description to work in the Subproject view.
subproject_description


oleg-nenashev:
Build descriptions may be pretty long. I doubt this single-line layout addresses all cases. Maybe makes sense to create a table. Once it's done, maybe makes sense to create BuildTableColumn extension point in Jenkins core and make the feature more extensible.

A short-term solution may be to explicitly limit the description length

svanoort:
Agree that this one makes more sense as a table... there's going to be a lot of overflow and *very* long text strings otherwise.

PR #94, opened 2015-10-27
convert parameters whose type is a subclass of StringParameterDefinition

jcarrothers-sap:
ParameterDefinition types which extend StringParameterDefinition still need
to be converted, when possible, in order to ensure correct handling of any
additional logic implemented by the sub-type.

This fix is motivated by a parameter type, in one of my own plugins, which inherits from StringParameterDefinition in order to also reproduce its environment value insertion for free.  However, builds which were triggered from this plugin were not correctly calling into the types additional build time logic.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/ProjectSpecificParameterValuesActionTransform.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/ProjectSpecificParameterValuesActionTransform.java
@@ -53,7 +53,7 @@ public class ProjectSpecificParameterValuesActionTransform implements ITransform
 
     private static boolean canConvert(ParameterDefinition def, ParameterValue v) {
         return def instanceof SimpleParameterDefinition &&
-            !(def instanceof StringParameterDefinition) &&
+            !(def.getClass().equals(StringParameterDefinition.class)) &&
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
I doubt the fix is correct, because many plugins seem to rely on the original behavior. Could you please create an issue in Jenkins JIRA and provide more details about the problem you face in your plugin?

jcarrothers-sap:
Do you have an example of a plugin that would be broken by this change?  This change actually expands the conditions under which a conversion is attempted, rather than limiting them further.  

The specific example I have stems from https://github.com/jcarrothers-sap/lockable-resources-plugin/blob/master/src/main/java/org/jenkins/plugins/lockableresources/RequiredResourcesParameterDefinition.java where I inherit from StringParameterDefinition in order to not have to duplicate the logic to inject the parameter value into the build environment.

jcarrothers-sap:
As requested: https://issues.jenkins-ci.org/browse/JENKINS-34009

jcarrothers-sap:
@oleg-nenashev Is there anything further I can do to help move this issue along?  We've been using a customized version of this plugin, in order to include this bug fix, in our production environment for months now without any issues.

oleg-nenashev:
Sorry, I have missed updates in this PR.

PR #95, opened 2015-11-20
JENKINS-17009: Parameters with newlines should be stored as TextParam…

zalan-axis:
Rebased centic9's patch from [1]: https://github.com/jenkinsci/parameterized-trigger-plugin/pull/36.

In the discussion in [1], it is suggested that this is NOT a bug in the parameterized-trigger-plugin.

I would like to argue that this is in fact a bug in parameterized-trigger-plugin since it makes assumptions about the specific ParameterValue type. It always assumes that if its a String than it must be a StringParameterValue, which is clearly wrong. TextParameterValue was invented for multi-line strings, and it cannot be disregarded, since Jenkins and other plugins (rebuild) use this type information to render the correct UI.

Without this patch the "Predefined parameters" and "Parameters from property file" are useless for multi-line strings. 

This is a trivial change, please be pragmatic about it.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameters.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/FileBuildParameters.java
@@ -19,6 +19,7 @@ import hudson.model.ParameterValue;
 import hudson.model.ParametersAction;
 import hudson.model.StringParameterValue;
 import hudson.model.TaskListener;
+import hudson.model.TextParameterValue;
 import hudson.util.FormValidation;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
:+1:

oleg-nenashev:
BTW I'm not sure it does not break existing use-cases. Maybe a global configuration option would be the best solution

zalan-axis:
Sure. I'll introduce global config paratemer for this. I assume it should be disabled by default for backward compatibility.

zalan-axis:
I've also patched FileBuildParameters because the same problem is found there as well.

In my second commit I've added a new "Don't trigger if any files are missing" checkbox to FileBuildParameters and PredefinedBuildParameters. This checkbox is by default off for backward compatibility as requested. The checkbox is hidden under "Advanced...".

As soon as this makes it into a release I will update the Job-DSL plugin as well to support this new option.

TWestling:
Is there a reason this pull request hasn't been merged?
Looking at the pull requests for the plugin, nothing has been merged since last year, has this plugin been orphaned?

TWestling:
@oleg-nenashev is there anything missing from this pull request or can it be merged?

oleg-nenashev:
Target release date - July 10

PR #96, opened 2015-12-14
[FIXED JENKINS-31727] - ConfigTrigger should be tolerant against Item.DISCOVER without Item.READ

oleg-nenashev:
In https://issues.jenkins-ci.org/browse/JENKINS-31727 the issue happens due to the explicit usage of Item.DISCOVER w/o Item.READ. Groovy UIs trigger AccessDeniedException, which finally leads to the UI corruption.

During the fix I've also adjusted implementations of other Getter methods in order to avoid failures.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
@@ -55,6 +55,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.concurrent.Future;
+import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
CC @svanoort

alecharp:
:bee:

Vlatombe:
:bee:

oleg-nenashev:
@reviewbybees done

svanoort:
:+1:  and merging as a pseudo-maintainer.  

Note that the try/catch use here to check for access is somewhat expensive due to creation of Exceptions.  For now that's fine, but if people report performance issues, we may want to look at a direct permission check for that, as cheaper than trying it and getting an exception.

PR #101, opened 2016-05-15
Fix Method not Allowed errors during svn checkout

ad22:
svn.jenkins-ci.org doesn't seem to support https checkout anymore. Building against the latest svn plugin doesn't work either. Changing the checkout protocol in SubversionRevisionBuildTriggerConfigTest to svn:// fixes it.  Local machine clone using svn, version 1.9.3 has the same behavior.


sschuberth:
This PR's description is a bit misleading as it simply is a copy of the second (and IMO less important) commit. The more important commit in this PR is " [JENKINS-11257]: Aborting the parent job will abort building/queued child jobs".

ad22:
This should ideally be two pull requests. I should have used two branches instead.  Created https://github.com/jenkinsci/parameterized-trigger-plugin/pull/103 for the SVN unit test errors and https://github.com/jenkinsci/parameterized-trigger-plugin/pull/104 for JENKINS-11257 (made the requested changes). Both of these are based off of the master branch instead of each other. Closing this request.

PR #102, opened 2016-05-17
Close file after reading.

lvotypko:
The stream is not closed after reading.


lvotypko:
I am not sure why the test failed, it seemed that problem was fixed before my change and it fails even if I remove my change....

ad22:
The svn unit test is failing. See https://github.com/jenkinsci/parameterized-trigger-plugin/pull/101 for a fix.

PR #103, opened 2016-05-18
Fix Method not Allowed errors during svn checkout

ad22:
svn.jenkins-ci.org doesn't seem to support https checkout anymore. Building against the latest svn plugin doesn't work either. Changing the checkout protocol in SubversionRevisionBuildTriggerConfigTest to svn:// fixes it. Local machine clone using svn, version 1.9.3 has the same behavior.


ad22:
Closing this as I think it was fixed in another commit

PR #106, opened 2016-09-01
Add possibility to set downstream projects for child projects.

lvotypko:


src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -96,7 +98,16 @@ public void buildDependencyGraph(AbstractProject owner, DependencyGraph graph) {
 		for (BuildTriggerConfig config : configs) {
 			List<AbstractProject> projectList = config.getProjectList(owner.getParent(), null);
 			for (AbstractProject project : projectList) {
-				ParameterizedDependency.add(owner, project, config, graph);
+                            if (config.triggerFromChildProjects() && owner instanceof ItemGroup) {
+                                ItemGroup parent = (ItemGroup) owner;
+                                for (AbstractProject child : (Collection<AbstractProject>)parent.getItems()) {
```
olivergondza: This cast is strictly speaking unsafe as there is no guarantee it contains `AbstractProject`s.

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerTest.java
```diff
@@ -0,0 +1,147 @@
+/*
+ * To change this template, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
olivergondza: To be deleted.

lvotypko:
Yes, sorry my IDE likes to add comments.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
```diff
@@ -96,7 +99,19 @@ public void buildDependencyGraph(AbstractProject owner, DependencyGraph graph) {
 		for (BuildTriggerConfig config : configs) {
 			List<AbstractProject> projectList = config.getProjectList(owner.getParent(), null);
 			for (AbstractProject project : projectList) {
-				ParameterizedDependency.add(owner, project, config, graph);
+                            if (config.triggerFromChildProjects() && owner instanceof ItemGroup) {
+                                ItemGroup<Item> parent = (ItemGroup) owner;
+                                for (Item item : parent.getItems()) {
+                                    if(item instanceof AbstractProject){
```
oleg-nenashev: I would vote against using AbstractProject in the new code, but it requires rework of `ParameterizedDependency`

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -116,6 +128,10 @@ public ResultCondition getCondition() {
 	public boolean getTriggerWithNoParameters() {
         return triggerWithNoParameters;
     }
+        
+    public boolean triggerFromChildProjects(){
```
oleg-nenashev: Replace by `isTriggerFromChildProjects` to follow the common naming convention

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -630,6 +646,10 @@ public String getDisplayName() {
             return Hudson.getInstance().<AbstractBuildParameterFactory,
               Descriptor<AbstractBuildParameterFactory>>getDescriptorList(AbstractBuildParameterFactory.class);
         }
+        
+        public boolean isItemGroup(AbstractProject project){
```
oleg-nenashev: Better to Restrict this method by `@Restricted(DoNotUse.class)` since it is only for UI

src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java
lvotypko: Thank you for all response, I put it into my code. I am afraid that only reworking ParametrizedDependency does not help.... https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/DependencyGraph.java#L394

lvotypko:
Oleg, this functionality should able to trigger the same job for every configuration for processing the results of particular configurations and uploading them to another server. It is true that we can use publishers but in case of matrix job, the publisher is used only for whole matrix - not for particular configuration. But we need to load results of every particular configuration separated.

PR #108, opened 2016-09-28
branch/show-subprojects-links-during-build

HeikoBoettger:
This branch changes the information presented in the upstream build while it's waiting for the triggered downstream projects to finish.

The goal is reached by using the QueueTaskFutures for polling status of the triggered builds. As soon as a build is started the id of the build (which is not available until it is queue) is registered in the BuildInfoExporterAction. 

The polling is done every 15 seconds.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
@@ -41,11 +41,84 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import jenkins.model.Jenkins;
+
+import org.apache.commons.beanutils.ConversionException;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
+import com.thoughtworks.xstream.converters.Converter;
+import com.thoughtworks.xstream.converters.MarshallingContext;
+import com.thoughtworks.xstream.converters.UnmarshallingContext;
+import com.thoughtworks.xstream.io.HierarchicalStreamReader;
+import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
+
 @ExportedBean
 public class BuildInfoExporterAction implements EnvironmentContributingAction {
+
+  public static class ConverterImpl implements Converter {
+
+    @Override
+    public boolean canConvert(Class type) {
+      return BuildInfoExporterAction.class.equals(type);
+    }
+
+    @Override
+    public void marshal(Object source, HierarchicalStreamWriter writer, MarshallingContext context) {
+      BuildInfoExporterAction action = (BuildInfoExporterAction) source;
+      writer.addAttribute("version", "3");
+      writer.startNode("builds");
+      context.convertAnother(action.builds);
+      writer.endNode();
+      writer.startNode("lastReference");
+      context.convertAnother(action.lastReference);
+      writer.endNode();
+    }
+
+    @Override
+    public Object unmarshal(HierarchicalStreamReader reader, UnmarshallingContext context) {
+      String versionStr = reader.getAttribute("version");
+      int version = (versionStr == null) ? 0 : Integer.valueOf(versionStr);
+      switch (version) {
+      case 0: {
+        // before 2.32
+        ArrayList<AbstractBuildReference> builds = new ArrayList<AbstractBuildReference>();
+        reader.moveDown();
+        if ("buildRefs".equals(reader.getNodeName())) {
+          // before 2.11 -> need to convert buildRefs to builds list
+          Map<String, List<BuildReference>> buildRefs = (Map<String, List<BuildReference>>) context.convertAnother(null,
+              Map.class);
+          for (List<BuildReference> buildRefsList : buildRefs.values()) {
+            builds.addAll(buildRefsList);
+          }
+        } else if ("builds".equals(reader.getNodeName())) {
+          builds.addAll((ArrayList<BuildReference>) context.convertAnother(null, ArrayList.class));
+        } else {
+          throw new ConversionException("unable to load data");
+        }
+        reader.moveUp();
+        reader.moveDown();
+        AbstractBuildReference lastReference = (AbstractBuildReference) context.convertAnother(null,
+            AbstractBuildReference.class);
+        reader.moveUp();
+        return new BuildInfoExporterAction(builds, lastReference);
+      }
+      case 3:
+        // since 2.32
+        reader.moveDown();
+        ArrayList<AbstractBuildReference> builds = (ArrayList<AbstractBuildReference>) context.convertAnother(null,
+            ArrayList.class);
+        reader.moveUp();
+        reader.moveDown();
+        AbstractBuildReference lastReference = (AbstractBuildReference) context.convertAnother(null,
+            AbstractBuildReference.class);
+        reader.moveUp();
+        return new BuildInfoExporterAction(builds, lastReference);
+      default:
+        throw new ConversionException("unsupported version");
+      }
+    }
+
+  }
 
   public static final String JOB_NAME_VARIABLE = "LAST_TRIGGERED_JOB_NAME";
   public static final String ALL_JOBS_NAME_VARIABLE = "TRIGGERED_JOB_NAMES";
@@ -58,23 +131,21 @@ public class BuildInfoExporterAction implements EnvironmentContributingAction {
   private transient String buildName;
   private transient int buildNumber;
 
-  // used in version =< 2.21.
-  // this is now migrated to this.builds.
-  private transient Map<String, List<AbstractBuildReference>> buildRefs;
-
   private List<AbstractBuildReference> builds;
 
-  // used in version =< 2.32.
-  // this is now migrated to this.lastReference2.
-  private transient BuildReference lastReference;
-  private AbstractBuildReference lastReference2;
+  private AbstractBuildReference lastReference;
 
   public BuildInfoExporterAction(AbstractBuildReference buildRef) {
     super();
 
     this.builds = new ArrayList<AbstractBuildReference>();
     addBuild(buildRef);
-    this.lastReference2 = buildRef;
+    this.lastReference = buildRef;
+  }
+
+  private BuildInfoExporterAction(List<AbstractBuildReference> builds, AbstractBuildReference lastReference) {
+    this.builds = builds;
+    this.lastReference = lastReference;
   }
 
   public BuildInfoExporterAction(AbstractBuild<?, ?> parentBuild, AbstractBuildReference buildRef) {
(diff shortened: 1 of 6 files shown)


HeikoBoettger:
I have added a small change since the I requested the pull. The logging during the poll is now going to a logger instead of the build's console.

kwhetstone:
Looks like there's still an error here and some conflicting files.  Also, could you add a test?

src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
```diff
@@ -129,6 +156,111 @@ public BuildReference(final String projectName) {
...
+      return buildResult;
+    }
+
+    @Override
+    public void update() {
+
+    }
+  }
+
+  public static class StaticBuildReference extends AbstractBuildReference {
```
kwhetstone: This looks identical to the BuildReference.  Is there a reason to use this over that?

src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
```diff
@@ -243,40 +380,48 @@ public void buildEnvVars(AbstractBuild<?, ?> build, EnvVars env) {
    * @return
    */
   public Object readResolve() {
-    if (this.lastReference == null) {
-      this.lastReference = new BuildReference(this.buildName, this.buildNumber, Result.NOT_BUILT);
+    if (this.lastReference2 == null) {
+      this.lastReference2 = this.lastReference;
```
kwhetstone: Does this ever have a chance to get set to anything?  It appears that all the changes now run through `lastReference2` so this potentially never gets set unless I'm missing something.

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -158,14 +280,42 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
...
+
+    
+    // Public but restricted so we can add tests without completely changing the
+    // tests package
+    @Restricted(value = org.kohsuke.accmod.restrictions.NoExternalUse.class)
+    public String getProjectListAsString(List<Job> projectList) {
+      return getProjectListAsString(new HashMap<Object, String>(), projectList);
+    }
+
+    public String getProjectListAsString(Map<Object, String> hyperlinkNotes, List<Job> projectList) {
```
kwhetstone: I think this also needs the `@Restricted(value=org.kohsuke.accmod.restrictions.NoExternalUse.class)` annotation.

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -158,14 +280,42 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
         return buildStepResult;
     }
 
-    // Public but restricted so we can add tests without completely changing the tests package
-    @Restricted(value=org.kohsuke.accmod.restrictions.NoExternalUse.class)
-    public String getProjectListAsString(List<Job> projectList){
+    private static String getOrCreateHyperlinkNote(Map<Object, String> hyperlinkNotes, Run run)
```
kwhetstone: +1 for this method cleaning up the rest of this class.

src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
HeikoBoettger: @kwhetstone 
Hi, thanks for the review.
I added a super class to the build reference and added two subclasses, one to get the information from the currently running build which needs to queried from Jenkins on every access (didn't found a way to do it different). The other one the StaticBuildReference is used when Jenkins serializes data between shutdown and restart which should do the same as BuildReference. However I kept the BuildReference since I didn't want to break the deserialization. It's a long time since I have used the built-in java serialization in combination with class renaming and I as far as I know it's impossible using a different class without being able to reimplement readClassDescriptor on the ObjectInputStream. However the ObjectInputStream is created by jenkins. Do you suggest just to keep the old name?

src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
HeikoBoettger: Again, this is a serialization thing. If lastReference was previously used it will be restored. The problem is that it seems not to be possible to just change lastReference to a more abstract type. At least it didn't worked for me but couldn't explain why it shouldn't.
Overall it probably would be better to use customized serialization (implementing readObject/writeObject) for all that stuff.  One has much more complete control over deserialization even after renaming fields and changing the internals. I will think about how to improve this.

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
HeikoBoettger: Thanks

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
HeikoBoettger: Yes, you are right.

HeikoBoettger: @kwhetstone 
Do you have a hint which existing testcase that is? I know how to unit test in general however I never did with jenkins. Any suggestions where to start?

src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
HeikoBoettger: @kwhetstone 
I made some tests an it seems like even the old code didn't work like expected. Do you know what the transient field buildRefs is supposed to do? I am asking because the resolveObject-method seems to expect that a transient field is restored if it wasn't transient before. However I made some tests and transient field aren't restored by java serialization nor by xstream.
transient before. However I made some tests and transient field aren't restored by java serialization nor by xstream.

HeikoBoettger: Forget my last comment, the problem is I tried to testing using plain xstream without using the jenkins specific xstream configuration. What a mess, depending on which documentation one is looking it is working or not. If you have a look at the BuildInfoExporterTest you will find a testcase called "testMigrateFrom221", this test loads an old 2.21 configuartion using the jenkins internal implementation. That test case will restore the lastReference although its set to transient since jenkins has configured xstream to do so. Since lastReference is only used in readResolve and the test is green this works. Probably you just overlooked that lastReference2 is used in addBuild, the constructor, getLastReference and here in readResolve as well. If the instance is serialized through xstream only lastReference2 is stored.
Do you know whether it's possible to switch to the ConverterImpl approach? This would be by far the cleaner way to separate the underlying storage format from the implementation.

HeikoBoettger: Details: https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility

src/main/java/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction.java
HeikoBoettger: @kwhetstone 
Please have a look whether the new solution looks better for you. I removed all the old fields and implemented a xstream-ConverterImpl handling the conversion.
Should I just leave the old name of BuildReference? Probably it's not worth renaming it just to have a "better" name describing that it's not changing once created. What do you think?

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -517,20 +526,111 @@ protected Future schedule(AbstractBuild<?, ?> build, final Job project, int quie
...
         // Trigger is not compatible with un-parameterized jobs
         return null;
     }
 
-    protected Future schedule(AbstractBuild<?, ?> build, Job project, List<Action> list) throws InterruptedException, IOException {
+    private QueueTaskFuture<? extends AbstractBuild> connectToExistingBuild(AbstractBuild<?, ?> currentBuild, Job<?, ? extends AbstractBuild> project, List<Action> actions) {
+        Run upstreamRoot = findUpstreamRoot(currentBuild, actions);
+
+        //always trigger the build if a user started the root build
+        Cause userIdCause = upstreamRoot.getCause(UserIdCause.class);
```
HeikoBoettger: @kwhetstone 
I need to remove this stuff. It's fits better to a new feature/pull request.
Is there any reason why the parameterized trigger plugin doesn't connect to jobs already build. My idea here is to over an option to allow using build result of job already finished and not trigger them again.

The scenario is as following: Having job A and B both depending on C. If C was build with the same parameters before and the result doesn't change if the parameters are the same (thats the tricky part), there is no need to rebuild it. There might be exceptional cases where a user might want to enforce building, so my first idea is to only reuse the result from a build finished in the past, if the build request wasn't started by a user (need to check the upper most upstream job). I think there should be some options (reduce number of builds by using results of existing build with same arguments + run a new build if uppermost upstream job was triggered by user) to keep compatibility and not breaking stuff for other users.

oleg-nenashev:
Needs re-review

kwhetstone:
I have to rereview.  This is such a huge change, it makes me slightly nervous, but I like the idea behind the change. @HeikoBoettger Your use case almost sounds like there should be a check on the back end of the C that would start A and B if there was actually a change.  A lot of the other features sound like future tickets (if they don't already exist) and new PRs.

oleg-nenashev:
@kwhetstone Any plans regarding the re-review?

HeikoBoettger:
@kwhetstone 
@oleg-nenashev 
There are a lot of changes now, I suggest to rework the solution. Also I discovered that there might be an issue when builds are cancelled or jenkins shutsdown while building. I also dislike my solution to create a bunch of additional threads just to make it possible to get notified when one of the futures is fulfilled. 

In the past I had a different approach, which was based on a queue listener triggering a loop to check all the future to find out which was finished. Unfortunately that didn't work because the futures are notified after the queue listener which caused a race condition, where after triggering the loop non of the futures was finished. It's a long time since I tried that and I wonder whether there is a to define a connection between a future and queue event so that it's clear which future/job was fulfilled.

HeikoBoettger:
@oleg-nenashev 
Strange things are happening, I rewrote the changes based on the latest master but it seems like the Pull Request still shows the old commits. Do you know how to solve this?

oleg-nenashev:
Hard to say without looking at the commit tree. Maybe merge/rebase were incorrect

orgads:
> Strange things are happening, I rewrote the changes based on the latest master but it seems like the Pull Request still shows the old commits. Do you know how to solve this?

Did you force-push to the branch? It looks like the branch on your clone is identical to this PR (latest commit on https://github.com/HeikoBoettger/parameterized-trigger-plugin/commits/branch/show-subprojects-links-during-build is 2466740)

PR #109, opened 2016-10-07
TriggerBuilder now implements DependencyDeclarer

TWestling:
Certain plugins doesn't work when triggering builds
of other projects as build steps, i.e. using TriggerBuilder.
When triggering as a post build step, i.e. using BuildTrigger,
this works. Two examples are the Junit plugin, when aggregating
test results from downstream builds and the downstream build view.

The reason is that BuildTrigger builds up a correct
DependencyGraph and TriggerBuilder does not.

This change remedies this by using the same mechanism as BuildTrigger.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
@@ -34,6 +34,7 @@ import hudson.model.AbstractProject;
 import hudson.model.Action;
 import hudson.model.BuildListener;
 import hudson.model.DependencyGraph;
+import hudson.model.TaskListener;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Builder;
@@ -177,7 +178,7 @@ public class TriggerBuilder extends Builder implements DependencyDeclarer{
         for (BuildTriggerConfig config : configs) {
             List<AbstractProject> projectList = config.getProjectList(owner.getParent(), null);
             for (AbstractProject project : projectList) {
-                ParameterizedDependency.add(owner, project, config, graph);
+                graph.addDependency(new TriggerBuilderDependency(owner, project, config));
             }
         }
     }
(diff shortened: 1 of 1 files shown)


GLundh:
Ping @olivergondza

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -158,6 +161,27 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
         return buildStepResult;
     }
 
+    private boolean canDeclare(AbstractProject owner) {
+        // In Hudson 1.341+ builds will be triggered via DependencyGraph
+        // Inner class added in Hudson 1.341
+        String ownerClassName = owner.getClass().getName();
+        return DependencyGraph.class.getClasses().length > 0
```
olivergondza: This seems to depend on an implementation detail.

olivergondza:
To my limited understanding, there is a build step to trigger a build and the downstream build might not observe the upstream build completed - that is an old struggle.

This fix, as I understand it, still triggers the build but it declares the dependency as well. Though, the dependency does not cause the downstream build to be triggered (again) but it is needed for clients that expect the dependency exists. This sort of makes sense to me.

I would say this needs unittest.

GLundh:
> This fix, as I understand it, still triggers the build but it declares the dependency as well. Though, the dependency does not cause the downstream build to be triggered (again) but it is needed for clients that expect the dependency exists. This sort of makes sense to me.

Yes, this is exactly how it works and enables plugins like "Downstream builds view plugin" to visualize the builds correctly.

> I would say this needs unittest.

Yes. Though I think existing tests are good enough to verify that the triggering is not being broken by this patch. But tests should be added to help verify the correctness of the provided DependencyGraph. Let me know if you don't think that is enough.

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
kwhetstone: In PR#112 I have to bump to version 1.642.3.  Would that help here?

GLundh:
Continued in #116 .

PR #110, opened 2016-12-23
Fix NPE in case of race condition

Jimilian:
NPE happens in rare case:
1) We have X projects
2) We disabled one of them
3) Trigger fired X-1 projects and filled map <projectName, executor>
4) Job was fixed, and we enable this job again
5) Trigger is waiting for all projects (now it's X projects again)
6) It tries to access job that we re-enabled -> it doesn't exist in map => NPE

How it looks like in logs:
```
Waiting for the completion of projectA
projectA #22344 completed. Result was SUCCESS
Waiting for the completion of projectB
ERROR: Build step failed with exception
java.lang.NullPointerException
Build step 'Trigger/call builds on other projects' marked build as failure
```

I failed to create test case for it :( .Because those two lines are too close to each other:
```
ListMultimap<Job, Future<Run>> futures = config.perform3(build, launcher, listener); // contains config.getJobs as well.
List<Job> projectList = config.getJobs(build.getRootBuild().getProject().getParent(), env);
```

In perfect world I would prefer different solution - take list of projects and provide this list to `perform3` step. But it needs much more changes and usually maintainers are not so happy to merge so complicated stuff :)

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
@@ -134,15 +134,17 @@ public class TriggerBuilder extends Builder {
                         }
                         for (Future<Run> future : futures.get(p)) {
                             try {
-                                listener.getLogger().println("Waiting for the completion of " + HyperlinkNote.encodeTo('/'+ p.getUrl(), p.getFullDisplayName()));
-                                Run b = future.get();
-                                listener.getLogger().println(HyperlinkNote.encodeTo('/'+ b.getUrl(), b.getFullDisplayName()) + " completed. Result was "+b.getResult());
-                                BuildInfoExporterAction.addBuildInfoExporterAction(build, b.getParent().getFullName(), b.getNumber(), b.getResult());
+                                if (future != null ) {
+                                    listener.getLogger().println("Waiting for the completion of " + HyperlinkNote.encodeTo('/'+ p.getUrl(), p.getFullDisplayName()));
+                                    Run b = future.get();
+                                    listener.getLogger().println(HyperlinkNote.encodeTo('/' + b.getUrl(), b.getFullDisplayName()) + " completed. Result was " + b.getResult());
+                                    BuildInfoExporterAction.addBuildInfoExporterAction(build, b.getParent().getFullName(), b.getNumber(), b.getResult());
 
-                                if(buildStepResult && config.getBlock().mapBuildStepResult(b.getResult())) {
-                                    build.setResult(config.getBlock().mapBuildResult(b.getResult()));
-                                } else {
-                                    buildStepResult = false;
+                                    if (buildStepResult && config.getBlock().mapBuildStepResult(b.getResult())) {
+                                        build.setResult(config.getBlock().mapBuildResult(b.getResult()));
+                                    } else {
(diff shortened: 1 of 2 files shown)


Jimilian:
@olivergondza, please, review :)

olivergondza:
> I failed to create test case for it :( .Because those two lines are too close to each other:

I do not follow. It should be possible to test the trigger build have not failed after downstream was re-enabled (and NPE is not logged).

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -134,15 +134,19 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
...
+                                    BuildInfoExporterAction.addBuildInfoExporterAction(build, b.getParent().getFullName(), b.getNumber(), b.getResult());
+
+                                    if (buildStepResult && config.getBlock().mapBuildStepResult(b.getResult())) {
+                                        build.setResult(config.getBlock().mapBuildResult(b.getResult()));
+                                    } else {
+                                        buildStepResult = false;
+                                    }
                                 } else {
-                                    buildStepResult = false;
+                                    listener.getLogger().println("Skipping " + HyperlinkNote.encodeTo('/'+ p.getUrl(), p.getFullDisplayName()) + ". The project was disabled.");
```
oleg-nenashev: It's only one of the potential null futures IIRC

Jimilian: Any suggestions what to put?
"The project was disabled or something went wrong"? :)

oleg-nenashev: Several cases:
- perform3 Javadoc: `// When a project is disabled or the configuration is not yet saved f will always be null and we're caught in a loop, therefore we need to check for it`
- BuildTriggerConfig#schedule:
 * job is not an instance of `ParameterizedJobMixIn.ParameterizedJob`
* `ParameterizedJobMixIn#schedule2()` returns null due to whatever other reason: Job is disabled, job is not buildable (e.g. triggering of Matrix Configuration), or Jenkins just does not like you (dispatcher extensions return null) :(

Jimilian: I can put ```The project was not triggered by some reason``` or 
```
The project was not triggered because one of possible cases happened: 
1) Project was disabled in moment then this build started 
2) Configuration of that project was not saved then this build started 
3) Project is not instance of "ParameterizedJobMixIn.ParameterizedJob"
4) Jenkins just does not like you (dispatcher extensions return null instead of running this project) :(
```
Or keep only first part `Skipping...`.

oleg-nenashev: "The project was not triggered by some reason" is fine for me. It's rather an RFE to Jenkins core in order to get a good CauseOfRejection as a response

Jimilian:
@oleg-nenashev, @olivergondza, can we merge this one as well? pleeeeaaaase :)

Jimilian:
@oleg-nenashev, @kwhetstone, @olivergondza can we merge it, please?

Jimilian:
@olivergondza Thanks!

PR #111, opened 2016-12-23
Fix extra comma error in list of projects

Jimilian:
Right now user can provide list of project with extra comma. I.e. `"projectA, "`
And verification passes.

But after execution user receives:
```
ERROR: Build aborted. Can't trigger undefined projects. 1 of the below project(s) can't be resolved:
 > 
 > projectA
Check your configuration!
```

It happens because in `TriggerBuilder.java` we have check like this:
```
// Get the actual defined projects
StringTokenizer tokenizer = new StringTokenizer(config.getProjects(env), ",");
...
if (tokenizer.countTokens() != projectList.size()) {
```

And as result number of tokens are bigger than number of projects (because we skipped empty one before). So, I adjust this check with the most earliest existed check I found.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
@@ -646,7 +646,7 @@ public class BuildTriggerConfig implements Describable<BuildTriggerConfig> {
             while(tokens.hasMoreTokens()) {
                 String projectName = tokens.nextToken().trim();
                 if (StringUtils.isBlank(projectName)) {
-                    return FormValidation.error(Messages.BuildTrigger_NoSuchProject(" ", "?"));
+                    return FormValidation.error("Blank project name in the list");
                 }
 
                 Item item = Jenkins.getInstance().getItem(projectName,project,Item.class); // only works after version 1.410
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
@@ -669,7 +669,6 @@ public class BuildTriggerConfig implements Describable<BuildTriggerConfig> {
                 }
 
                 hasProjects = true;
-
             }
             if (!hasProjects) {
             	return FormValidation.error(Messages.BuildTrigger_NoProjectSpecified());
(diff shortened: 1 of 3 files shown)


Jimilian:
@olivergondza, please, review :)

olivergondza:
Looks, good to me (https://github.com/jenkinsci/parameterized-trigger-plugin/pull/111/files?w=1) but is missing tests.

Jimilian:
@kwhetstone, can you merge, please?

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -645,28 +645,30 @@ public FormValidation doCheckProjects(@AncestorInPath Job<?,?> project, @QueryPa
...
-                    if(item==null){
-                        Item nearest = Items.findNearest(Job.class, projectName, Jenkins.getInstance());
-                        String alternative = nearest != null ? nearest.getRelativeNameFrom(project) : "?";
-                        return FormValidation.error(Messages.BuildTrigger_NoSuchProject(projectName, alternative));
-                    }
-                    if(!(item instanceof Job) || !(item instanceof ParameterizedJobMixIn.ParameterizedJob)) {
-                        return FormValidation.error(Messages.BuildTrigger_NotBuildable(projectName));
-                    }
+                if (StringUtils.isBlank(projectName)) {
+                    return FormValidation.error(Messages.BuildTrigger_NoSuchProject(" ", "?"));
```
oleg-nenashev: I would rather return another warning ("Blank job name in the list")

Jimilian: All other messages are using `hudson.tasks.Messages` from `core`. And most of them are localized. Is it ok to put custom message? It could break user's expectations.

oleg-nenashev: It's a new message, so IMHO it's fine. Jenkins core just skips such empty lines without any message:

```java
public static <T extends Item> List<T> fromNameList(ItemGroup context, @Nonnull String list, @Nonnull Class<T> type) {
        final Jenkins jenkins = Jenkins.getInstance();
        
        List<T> r = new ArrayList<T>();
        if (jenkins == null) {
            return r;
        }
        
        StringTokenizer tokens = new StringTokenizer(list,",");
        while(tokens.hasMoreTokens()) {
            String fullName = tokens.nextToken().trim();
            if (StringUtils.isNotEmpty(fullName)) {
                T item = jenkins.getItem(fullName, context, type);
                if(item!=null)
                    r.add(item);
            }
        }
        return r;
    }
```

PR #112, opened 2017-03-03
plugin compat tester fixes

kwhetstone:
With the new changes to the plugin compatibility tester, a few of this plugin's tests fail.  The plugin compat tester tests a plugin against the newest version of it's dependencies as if it's been installed on the newest version of Jenkins.  This led to some failures even though when running tests on the current Jenkins version were no issues.

Currently this depends on the 1.x line of the workflow plugins.  Since there were significant changes with this plugin upon the 2.x release, I updated them and included new plugins that replicate the old functionality.  Also, I bumped the plugin-pom version as well as matirx-auth to get out of the "break-cycles" issue.  I had to bump the jenkins.version to 1.638 for missing pieces for the new workflow plugins.  That's the lowest version that I could move to.  However, it might be worth going to the next LTS for stability.

I also had to address the many SECURITY-170 violations that underpined many of the tests.  SECURITY-170 really targeted this plugin, so I had to change the tests to reflect that.  This also makes the tests future proof when we actually depend on versions above 2.3.

 I have run all the parameterized-trigger tests using mvn test as well as through the plugin compat tester, and there are now no errors.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,8 +152,7 @@
     </pluginRepositories>
     
     <properties>
-      <jenkins.version>1.638</jenkins.version>
-      <java.level>6</java.level>
+      <jenkins.version>1.642.3</jenkins.version>
       <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
 </project>  
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,7 @@
         <dependency>
           <groupId>org.jenkins-ci.plugins</groupId>
           <artifactId>matrix-auth</artifactId>
-	      <version>1.3.2</version>
+	      <version>1.4</version>
 	      <scope>test</scope>
         </dependency>
         <dependency>
@@ -128,12 +128,6 @@
 	            <artifactId>hamcrest-core</artifactId>
              </exclusion>
           </exclusions>
-        </dependency>
-        <dependency>
-          <groupId>org.jenkins-ci.plugins</groupId>
-          <artifactId>cloudbees-folder</artifactId>
-	      <version>5.18</version>
-	      <scope>test</scope>
         </dependency>
     </dependencies>
 
@@ -158,7 +152,7 @@
     </pluginRepositories>
     
     <properties>
-      <jenkins.version>1.637</jenkins.version>
+      <jenkins.version>1.638</jenkins.version>
       <java.level>6</java.level>
       <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.7</version>
+        <version>2.23</version>
     </parent>
     
     <licenses>
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.10</version>
+            <version>2.9</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -134,8 +152,7 @@
     </pluginRepositories>
     
     <properties>
-      <workflow.version>1.4</workflow.version>
-      <jenkins.version>1.596.1</jenkins.version>
+      <jenkins.version>1.638</jenkins.version>
       <java.level>6</java.level>
```
jglick: Delete; that version of Jenkins does not run on 6.

pom.xml
```diff
@@ -78,25 +78,43 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>${workflow.version}</version>
+            <version>2.9</version>
```
jglick: All 2.x Pipeline plugins require at least 1.642.3.

kwhetstone:
Bumped to 1.642.3.

src/test/java/hudson/plugins/parameterizedtrigger/test/DefaultParametersTest.java
```diff
@@ -61,28 +61,38 @@ public void test() throws Exception {
...
-		assertNotNull("projectB should be triggered: " + log, q);
-		q.getFuture().get();
-		assertNotNull("builder should record environment", builder.getEnvVars());
-		assertEquals("value1", builder.getEnvVars().get("KEY1"));
-		assertEquals("value2", builder.getEnvVars().get("KEY2"));
-		assertEquals("value3", builder.getEnvVars().get("KEY3"));
+        try {
+            // SECURITY-170: this is needed for tests down the line.
+            //System.setProperty(ParametersAction.KEEP_UNDEFINED_PARAMETERS_SYSTEM_PROPERTY_NAME, "true");
+            System.setProperty("hudson.model.ParametersAction.keepUndefinedParameters", "true");
```
jglick: Why not just define the parameters?

src/test/java/hudson/plugins/parameterizedtrigger/test/FileBuildParameterFactoryTest.java
```diff
@@ -245,22 +257,31 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
             }
         });
 
-        // add Trigger builder, with file parameter factory
-        projectA.getBuildersList().add(createTriggerBuilder(projectB, NoFilesFoundEnum.SKIP, "UTF-8"));
+        // SECURITY-170: need to allow multibyte params that can't be traditionally declared.
```
jglick: Why not?

src/test/java/hudson/plugins/parameterizedtrigger/test/TriggerBuilderTest.java
```diff
@@ -268,7 +268,7 @@ public void testBlockingTriggerWithDisabledProjects() throws Exception {
     /** Verify that workflow build can be triggered */
     @Test
     public void testTriggerWithWorkflow() throws Exception {
-        WorkflowJob p = (WorkflowJob) Jenkins.getInstance().createProject(WorkflowJob.class, "project1");
+        WorkflowJob p = (WorkflowJob) r.createProject(WorkflowJob.class, "project1");
```
jglick: Superfluous cast.

src/test/java/hudson/plugins/parameterizedtrigger/test/FileBuildParameterFactoryTest.java
kwhetstone: When I tested locally, I couldn't declare the multibyte params causing this test to fail.

jglick: Why not? Is your IDE not honoring the POM declaration of UTF-8 sources or something?

kwhetstone: I had coped things over in vi, which I thought would have made the situation easier, but didn't seem to do anything.  At least the global param exists.

jglick: > I had coped things over in vi

Srsly? :-) Well I know at least NetBeans correctly loads & saves all project sources in the encoding declared in the POM…would hope IDEA does the same.

PR #113, opened 2017-06-13
[JENKINS-41631] Demonstration of POM update

jglick:
Downstream of https://github.com/jenkinsci/plugin-pom/pull/67.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.31-20170612.235547-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/67 -->
+        <version>2.31</version>
         <relativePath/>
     </parent>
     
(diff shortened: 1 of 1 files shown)


jglick:
Was building with JDK 7.

pom.xml
```diff
@@ -4,7 +4,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.23</version>
+        <version>2.31-20170612.235547-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/67 -->
```
kwhetstone: If you don't need this right away, do you want to wait until the next version is cut?  This plugin isn't released frequently, so it might be worth waiting until the plugin-pom is updated.

jglick: Certainly I was not proposing a merge unless and until the dependency is merged and released. That just applies to any use of snapshots.

Really this PR is not intended to accomplish anything in particular except to demonstrate that the upstream PR is consumable.

oleg-nenashev:
Target release date - July 10

jglick:
@reviewbybees done

PR #115, opened 2017-07-12
[JENKINS-45471] - Generalize handling of scheduling errors and prevent NPEs in perform3()

oleg-nenashev:
The change polishes #114 and adds check of another NPE risk in `BuildTriggerConfig#perform3()`. Will add tests later... really...

https://issues.jenkins-ci.org/browse/JENKINS-45471

@reviewbybees @fbelzunc

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
@@ -419,7 +419,7 @@ public class BuildTriggerConfig implements Describable<BuildTriggerConfig> {
         listener.error("Skipping " + jobToTrigger.getFullName() + "...");
         if (LOGGER.isLoggable(Level.CONFIG)) {
             String message = String.format("Cannot schedule project %s. Job type is not parameterized, "
-                + "or there is no Item/Build permission for the current authentication %s. "
+                + "or there is no Job/Build permission for the current authentication %s. "
                 + "Skipping...", jobToTrigger, Jenkins.getAuthentication().getName());
              LOGGER.log(Level.CONFIG, String.format("%s: %s", run, message), 
                      new UnsupportedOperationException("Cannot schedule job " + jobToTrigger.getFullName()));
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
@@ -401,7 +401,7 @@ public class BuildTriggerConfig implements Describable<BuildTriggerConfig> {
                         if (future != null) {
                             futures.add(future);
                         } else {
-                            LOGGER.log(Level.FINE, "The schedule for project {0} and build {1} failed due either security reasons or the trigger is not compatible with un-parameterized jobs", new Object[]{project.getFullName(), build.number});
+                            reportSchedulingError(build, project, listener);
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -412,9 +412,20 @@ private static void resolveProject(AbstractBuild build, SubProjectData subProjec
...
-	}
-
+    }
 
+    private void reportSchedulingError(@Nonnull Run<?, ?> run, @Nonnull Job<?, ?> jobToTrigger, @Nonnull BuildListener listener) {
+        // Do not print details to Build Listener, they have been reported previously in #canTriggerProject()
+        listener.error("Skipping " + jobToTrigger.getFullName() + "...");
+        if (LOGGER.isLoggable(Level.CONFIG)) {
+            String message = String.format("Cannot schedule project %s. Job type is not parameterized, "
+                + "or there is no Item/Build permission for the current authentication %s. "
```
jglick: As I have mentioned previously, `Item/Build` is nonexistent. The Java constant is `Item.BUILD`. The permission as displayed in the UI is `Job/Build`.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -412,9 +412,20 @@ private static void resolveProject(AbstractBuild build, SubProjectData subProjec
...
+                + "Skipping...", jobToTrigger, Jenkins.getAuthentication().getName());
+             LOGGER.log(Level.CONFIG, String.format("%s: %s", run, message), 
+                     new UnsupportedOperationException("Cannot schedule job " + jobToTrigger.getFullName()));
```
jglick: Is a stack trace really necessary here? I do not see why.

src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
```diff
@@ -412,9 +412,20 @@ private static void resolveProject(AbstractBuild build, SubProjectData subProjec
...
 		}
         return Collections.emptyList();
-	}
-
+    }
 
+    private void reportSchedulingError(@Nonnull Run<?, ?> run, @Nonnull Job<?, ?> jobToTrigger, @Nonnull BuildListener listener) {
+        // Do not print details to Build Listener, they have been reported previously in #canTriggerProject()
+        listener.error("Skipping " + jobToTrigger.getFullName() + "...");
+        if (LOGGER.isLoggable(Level.CONFIG)) {
```
jglick: `BuildTrigger` and `ReverseBuildTrigger` print similar messages at `WARNING`. Is there a reason you are suppressing this message from logs by default? It is unlikely an admin will ever see it—by the time they figure out they need to be turning on this logger, they will already know about the problem and the fix. The point is to guide admins to help their users address configuration problems.

oleg-nenashev: @jglick ADMIN already gets a WARNING here: https://github.com/jenkinsci/parameterized-trigger-plugin/blob/master/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java#L571-L582

This logging with Exception exists for extra diagnostics purpose only

jglick: OK…confusing but it sounds like this is just for debugging?

oleg-nenashev: Yep. Will add more comments during the logging update in the plugin

PR #121, opened 2018-04-23
Make multiline parameter when newlines are present with indirect value

raspy:
When user defines a multiline parameter:

image

then passes this parameter downstream as B=$X:

image

then the parameter information in the downstream job is not rendered as multiline:

image

Direct root cause is the check for embedded newline character before expanding parameter value. This pull request fixes the behavior by expanding the string before newline check. I have also added a unit test for this scenario.

As a side note: I am not sure why there is a checkbox for the user to create TextParameterValue instead of StringParameterValue. I believe that it should always be rendered as TextParameterValue; there is no point to present this option to user, especially since this is implementation detail and is totally meaningless for the end user.

### DIFF
--- a/src/test/java/hudson/plugins/parameterizedtrigger/test/PredefinedPropertiesBuildTriggerConfigTest.java
+++ b/src/test/java/hudson/plugins/parameterizedtrigger/test/PredefinedPropertiesBuildTriggerConfigTest.java
@@ -2,6 +2,7 @@
  * The MIT License
  *
  * Copyright (c) 2004-2010, Sun Microsystems, Inc., Kohsuke Kawaguchi
+ * Copyright (c) 2018 Motorola Solutions, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


raspy:
@kwhetstone could you please help to get this change reviewed?

raspy:
@kwhetstone could you please help to get this change integrated?

src/test/java/hudson/plugins/parameterizedtrigger/test/PredefinedPropertiesBuildTriggerConfigTest.java
```diff
@@ -2,6 +2,7 @@
  * The MIT License
  *
  * Copyright (c) 2004-2010, Sun Microsystems, Inc., Kohsuke Kawaguchi
+ * Copyright (c) 2018 Motorola Solutions, Inc.
```
kwhetstone: I'm not too comfortable with adding a copyright for a company for an OSS plugin.  I'm fine with even just removing the Sun copyright line as well.

raspy: Thank you for the review. Unfortunately our internal requirements oblige me to put this in to be allowed to submit the change back to the upstream. However it only requires to do so if there is already a copyright, so I believe if all the copyrights are removed altogether then it would be fine as well.

raspy:
@kwhetstone @jglick @oleg-nenashev I have rebased the PR to the newest version. Could you please help to have it included in official release?

jglick:
I am not a maintainer and have no interest in this pre-Pipeline plugin, sorry.

PR #122, opened 2018-07-18
add a period to the console every 5 seconds to avoid timeouts

smccarthy:


smccarthy:
@kwhetstone Hi. Can you possible review my PR or let me know someone else who might be able to?

Thank you!

smccarthy:
@oleg-nenashev Here is the one I was talking about in the irc channel.

smccarthy:
@kwhetstone I am not sure. I have never used java's Future's before. However, I would think this wouldn't make it more of a problem.

ad22:
My 2c: This seems like intentionally breaking the timeout functionality. We use this plugin extensively and have use cases where we want the parent build to time out instead of waiting indefinitely for child jobs to finish.

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -144,6 +145,11 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
                             try {
                                 if (future != null ) {
                                     listener.getLogger().println("Waiting for the completion of " + HyperlinkNote.encodeTo('/'+ p.getUrl(), p.getFullDisplayName()));
+                                    while(!future.isDone()) {
+                                        listener.getLogger().print(".");
+                                        Thread.sleep(SLEEP_MILLISECONDS);
```
ad22: I think you will need to catch InterruptedException here and re-throw it. If the parent thread is interrupted (job aborted) during a blocking operation, it needs to be propagated, otherwise the job may never terminate.

smccarthy:
@ad22 Wouldn't the timeout be on the child job, and not on the parent job ?

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
smccarthy: Ok thanks, I will look into this.

ad22:
@smccarthy In a scenario where something/someone is waiting for feedback from the parent jenkins job - suppose a child job is blocked due to an agent/label that does not exist/permanently offline, you would want that parent job to abort after a fixed time instead of waiting indefinitely.

PR #125, opened 2019-05-15
[JENKINS-57442]: print link to running job

timarmstrong:
This makes it much easier to find the child job while it
is running and not yet finished.

This is implemented with QueueTaskFuture's waitForStart()
method that returns the job once it starts running. The Future
waited on was already an instance of QueueTaskFuture, but
I needed to plumb through the type instead of the Future
interface.

Testing:
* Added checks to a couple of unit tests for the new log lines.
* Ran unit tests.
* Tested plugin on a production jenkins instance.


timarmstrong:
@kwhetstone you seem to be one of the reviewers for this project

timarmstrong:
@kwhetstone thanks! Do you have permissions to merge this?

PR #128, opened 2019-09-01
Feature: System property ptp.disableActionViews disables summary

GLundh:
Set Java property "ptp.disableActionViews" (typically Launch Jenkins
with -Dptp.disableActionViews) to stop the listing of "Sub Projects"
on the build and project main pages. This can be of use when other
plugins are used to render the upstream/downstream flow and there is
reduntant information displayed about the triggered builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,6 @@ This plugin lets you trigger new builds when your build has completed, with vari
 
 See the documentation and release notes at [Parameterized Trigger Plugin](https://wiki.jenkins-ci.org/display/JENKINS/Parameterized+Trigger+Plugin) on the Jenkins Wiki for more information.
 
-
-
-
+Note: Set Java property "ptp.disableActionViews" (typically Launch Jenkins with -Dptp.disableActionViews) to stop the listing of
+"Sub Projects" on the build and project main pages. This can be of use when other plugins are used to render the upstream/downstream
+flow and there is reduntant information displayed about the triggered builds.
--- a/src/main/resources/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction/summary.groovy
+++ b/src/main/resources/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction/summary.groovy
@@ -1,5 +1,11 @@
 package hudson.plugins.parameterizedtrigger.BuildInfoExporterAction
 
+import static hudson.plugins.parameterizedtrigger.ParameterizedTriggerUtils.DISABLE_ACTION_VIEWS_KEY
+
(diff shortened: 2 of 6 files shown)


GLundh:
Ping @jonsten

src/main/resources/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction/summary.groovy
```diff
@@ -1,5 +1,8 @@
 package hudson.plugins.parameterizedtrigger.BuildInfoExporterAction
 
+if (System.getProperty("ptp.disableActionViews") != null) {
```
jonsten: Considering that this check is done four times, how about extracting the logic into a boolean static helper method, maybe in ParameterizedTriggerUtils?

README.md
```diff
@@ -4,6 +4,6 @@ This plugin lets you trigger new builds when your build has completed, with vari
 
 See the documentation and release notes at [Parameterized Trigger Plugin](https://wiki.jenkins-ci.org/display/JENKINS/Parameterized+Trigger+Plugin) on the Jenkins Wiki for more information.
 
-
-
-
+Note: Set Java property "ptp.disableActionViews" (typically Launch Jenkins with -Dptp.disableActionViews) to stop the listing of
```
jonsten: Isn't convention that the property starts with the package id, i.e. in this case hudson.plugins.parameterizedtrigger.disableActionViews?

src/main/resources/hudson/plugins/parameterizedtrigger/BuildInfoExporterAction/summary.groovy
GLundh: I considered it, but decided against it since it would increase number of lines with approx 50% without increasing readability. Though I will lift out the property key to a constant to make adjustments easier.

README.md
GLundh: Yes, in Jenkins core it is, but plugins generally seems to want to keep it shorter (like YABV).

Check out the results here: https://github.com/search?q=org%3Ajenkinsci+System.getProperty+repo%3Aplugin&type=Code

Checking the first few pages, the most common way to define system properties for plugins seems to be the usage of a short plugin alias + "." + a verb or noun.

With that in mind, I'm certainly fine with using a package name if that is still what you prefer.

GLundh:
@kwhetstone : Would you mind taking a look and approving if you are fine with the code?

GLundh:
Ping @ikedam @oleg-nenashev @olivergondza @svanoort @wolfs

oleg-nenashev:
It was released in 2.36

PR #130, opened 2019-10-17
Update dependency on the Jenkins core and the Promoted Builds plugin

froque:


Simmay93:
Will this be reviewed in near future? Our promotions do not work because of this error.

JoaoPPinto:
@oleg-nenashev any chance you will review this in the near future?

oleg-nenashev:
I doubt I am even the plugin's maintainer per-se. But I will try to cut a release if @kwhetstone is fine with that

froque:
> It might stabilize one particular plugin, but it will push all users of Promoted Builds to use the version which causes regressions in other plugins. AI advice against doing such update until a proper fix is implemented in Promoted Builds

This pull request was meant to make parameterized-trigger work with promoted-builds. I also made a PR in publish-over-plugin (https://github.com/jenkinsci/publish-over-plugin/pull/10). 

If upgrading all dependant plugins is not the way to go, please elaborate on what you mean by a proper fix in Promoted Builds. I am willing spend some time with it, but can't see a proper fix without going back to bug [JENKINS-59600](https://issues.jenkins-ci.org/browse/JENKINS-59600)

oleg-nenashev:
From what I see we can revert the core version to 2.121.1 in the plugin and
then rollback the 3.4 fix. It would postpone the issue afaict

What we could do in longer term...

1) Accept the state of Promoted Builds API, release 4.0
2) Create a new plugin (e.g. Promotion API like it was planned for Pipeline
support OR Promoted Builds Migration API). This plugin should support both
3.2 and 3.4+
3) Update plugins to use the new API instead of the plugin directly

oleg-nenashev:
I will try to create a bugfix by Monday EoD. Should have some time at the
Jenkins hackfest.

oleg-nenashev:
https://github.com/jenkinsci/promoted-builds-plugin/pull/144 . It was tested manually, and it will be binary compatible with this fix

froque:
@oleg-nenashev  bumped promoted-builds to 3.5

PR #135, opened 2019-12-10
[JENKINS-57507] Recover from orphaned labels

elliot-nelson:
### SUMMARY

Proposed fix for https://issues.jenkins-ci.org/browse/JENKINS-57507.

### DETAILS

If the parameterized trigger plugin triggers a build, using a `NodeAction` to refer to a specific label or machine (self-label), it attaches the NodeAction directly to a Label object.

Normally this is fine, but if the machine disconnects at the wrong moment, you can end up with an "orphaned label".  It becomes orphaned because Jenkins periodically cleans up Labels that have no connected nodes, and then recreates a new one for the node when it reconnects.  When this happens, if you retain a reference to the individual Label object, that Label will never see any newly connected nodes.

The fix is for the `getAssignedLabel` method to always return the "canonical" label for for the given label expression, by asking the Jenkins instance for it, rather than returning our (potentially stale) Label reference.

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/NodeAction.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/NodeAction.java
@@ -34,23 +34,29 @@ import hudson.model.queue.SubTask;
 import java.util.List;
 
 /**
- * {@link Action} that restricts the job to a particular node 
+ * {@link Action} that restricts the job to a particular node
  * when a project is scheduled
  * Will cause a unique build for each different node if a job is already queued.
- * 
+ *
  * @author Chris Johnson
  */
 public class NodeAction extends InvisibleAction implements LabelAssignmentAction, Queue.QueueAction, BuildBadgeAction {
 	private final Label nodeLabel;
-	
+
 	public NodeAction(Label nodeLabel) {
 		this.nodeLabel = nodeLabel;
 	}
 
 	public Label getAssignedLabel(SubTask task) {
-		return nodeLabel;
+		if (nodeLabel != null) {
+			// Ensure that we are returning the canonical Label for this expression,
+			// according to the Jenkins instance.
+			return Label.get(nodeLabel.getName());
+		} else {
+			return nodeLabel;
+		}
 	}
-	
+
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/parameterizedtrigger/NodeAction.java
```diff
@@ -34,23 +34,29 @@
...
- * {@link Action} that restricts the job to a particular node 
+ * {@link Action} that restricts the job to a particular node
  * when a project is scheduled
  * Will cause a unique build for each different node if a job is already queued.
- * 
+ *
  * @author Chris Johnson
  */
 public class NodeAction extends InvisibleAction implements LabelAssignmentAction, Queue.QueueAction, BuildBadgeAction {
 	private final Label nodeLabel;
```
oleg-nenashev: SpotBugs is disabled in the plugin, but it would be a good practice to have `CheckForNull` annotations

elliot-nelson:
> This PR rather indicates an issue in the Jenkins core IMHO, it should be more robust. `Label.get(nodeLabel.getName());` will led to some performance degradation if there are many entries in the queue, but I doubt the overhead is significant enough to require a patch.
> 
> CC @jenkinsci/core just in case there are performance considerations

This makes sense; I considered this a bug in `NodeAction` because I couldn't find examples of long-term holding onto a Label (as opposed to maybe just eliminating that field and using a string for the label expression).

But if Jenkins core were changed so it could detect and automatically fix the "label resetting" issue I described in that ticket, maybe no change would be necessary here.

daniel-beck:
@olamy Rather than spam everyone with this, have you considered maintaining the plugin instead?

olamy:
@daniel-beck maybe 365days is not long enough. but too old PR not interest should be closed ( some are from 2013 with last activity 2015) what is the point having. At least we ping contributors of those PRs if still interested.
You can choose to receive notifications only when participating or mentioned.

daniel-beck:
@olamy Then do that, explaining you have no interest in the PR. Don't hide behind a bot that just spams everyone who's ever been mentioned.

PR #138, opened 2019-12-19
fix ArrayIndexOutOfBoundsException

yuiung:
when using dynamic parameter trigger and the count of dynamic downstream projects is greater than 2,there is an ArrayIndexOutOfBoundsException.the VAR  doesn't seem to work,so please cut it!


olamy:
duplicate of #139

PR #139, opened 2019-12-20
Avoid array index  out of bound

ahahxof:
I have a project that use variables to trigger parameterized build. 
I.e.g Instead of using job name, it is $jobs where jobs=job1,job2,job3
In this case, the existing code throws array index out of bound exception. 
Any reason code is iterating projects array and test vars[i]?
Also I don't think testing vars[i].trim().contains("$") is needed


olamy:
@ahahxof as you know some use case maybe you can add a test?

PR #161, opened 2020-12-31
Fix Null pointer exceptions

res0nance:
Should also prevent https://issues.jenkins.io/browse/JENKINS-64502?focusedCommentId=402814&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-402814

Drops the number of spotbugs to 4 from 10

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
@@ -153,9 +153,11 @@ public class TriggerBuilder extends Builder implements DependencyDeclarer {
                                     listener.getLogger().println(HyperlinkNote.encodeTo('/' + completedRun.getUrl(), completedRun.getFullDisplayName()) + " completed. Result was " + completedRun.getResult());
                                     BuildInfoExporterAction.addBuildInfoExporterAction(build, completedRun.getParent().getFullName(), completedRun.getNumber(), completedRun.getResult());
 
-                                    Result r = config.getBlock().mapBuildResult(completedRun.getResult());
-                                    if (buildStepResult && config.getBlock().mapBuildStepResult(completedRun.getResult()) && r != null) {
-                                        build.setResult(r);
+                                    if (buildStepResult && config.getBlock().mapBuildStepResult(completedRun.getResult())) {
+                                        Result r = config.getBlock().mapBuildResult(completedRun.getResult());
+                                        if (r != null) {
+                                            build.setResult(r);
+                                        }
                                     } else {
                                         buildStepResult = false;
                                     }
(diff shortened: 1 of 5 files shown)


res0nance:
CC: @olamy
Probably should add this bit if releasing as this takes care of some edge cases causing null pointers to be produced

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -133,9 +134,11 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
                     for (Job p : projectList) {
                         //handle non-buildable projects
                         if(!config.canBeScheduled(p)){
+                            User user = User.current();
+                            String userName = user != null ? ModelHyperlinkNote.encodeTo(user) : "unknown";
                             listener.getLogger().println("Skipping " + HyperlinkNote.encodeTo('/'+ p.getUrl(), p.getFullDisplayName()) + 
                                     ". The project is either disabled,"
-                                    + " or the authenticated user " + ModelHyperlinkNote.encodeTo(User.current()) + " has no Item.BUILD permissions,"
+                                    + " or the authenticated user " + userName + " has no Item.BUILD permissions,"
```
oleg-nenashev, line 141: "the authenticated user unknown has no ..." is a bit odd, but should be ok

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
```diff
@@ -151,7 +154,10 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
                                     BuildInfoExporterAction.addBuildInfoExporterAction(build, completedRun.getParent().getFullName(), completedRun.getNumber(), completedRun.getResult());
 
                                     if (buildStepResult && config.getBlock().mapBuildStepResult(completedRun.getResult())) {
-                                        build.setResult(config.getBlock().mapBuildResult(completedRun.getResult()));
+                                        Result r = config.getBlock().mapBuildResult(completedRun.getResult());
+                                        if (r != null) {
```
oleg-nenashev, line 158: `r == null ` when the run is not finished.
Maybe it makes sense to add a system warning for the edge case

oleg-nenashev (approved):

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
res0nance: It should be `the authenticated user unknown has no Item.BUILD`

src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
res0nance: Sure I can add it in case there is some underlying issue here

PR #169, opened 2021-02-17
More logging for BuildTriggerTest.testParentProjectTrigger

jglick:
Saw a `plugin-compat-tester` failure

```
java.lang.AssertionError: Downstream job should be triggered
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertNotNull(Assert.java:713)
	at hudson.plugins.parameterizedtrigger.test.BuildTriggerTest.testParentProjectTrigger(BuildTriggerTest.java:65)
```

Guessing this was a race condition but hard to know since there are no log messages while the test runs:

```
…
0.793 [id=…]	INFO	jenkins.InitReactorRunner$1#onAttained: Completed initialization
4.662 [id=…]	INFO	jenkins.model.Jenkins#cleanUp: Stopping Jenkins
…
```

This patch helps some, though if https://github.com/jenkinsci/jenkins/pull/4345#discussion_r348496722 had been accepted it would have helped more.

### DIFF
--- a/src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerTest.java
+++ b/src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerTest.java
@@ -17,6 +17,7 @@ import hudson.model.FreeStyleBuild;
 import hudson.model.FreeStyleProject;
 import hudson.model.ParameterDefinition;
 import hudson.model.ParametersDefinitionProperty;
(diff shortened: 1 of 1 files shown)


raul-arabaolaza (approved):
FWIW I have suffered similar issues

PR #235, opened 2022-01-17
Add spotbugs exclusions

dheerajodha:
This PR simply updates the parent POM of this plugin from 4.31 to 4.33.


src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -24,19 +26,24 @@
...
+    public BlockableBuildTriggerConfig(String projects, BlockingBehaviour block,
+            List<AbstractBuildParameters> configs) {
         super(projects, ResultCondition.ALWAYS, false, configs);
         this.block = block;
     }
 
     @DataBoundConstructor
-    public BlockableBuildTriggerConfig(String projects, BlockingBehaviour block, List<AbstractBuildParameterFactory> configFactories,List<AbstractBuildParameters> configs) {
+    public BlockableBuildTriggerConfig(String projects, BlockingBehaviour block,
+            List<AbstractBuildParameterFactory> configFactories, List<AbstractBuildParameters> configs) {
```
jglick, line 46: By the way please avoid gratuitous diff hunks, to make reviewing real changes easier.

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -46,46 +53,57 @@ public BlockingBehaviour getBlock() {
...
     @Override
-    public ListMultimap<AbstractProject, QueueTaskFuture<AbstractBuild>> perform2(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-        ListMultimap<AbstractProject, QueueTaskFuture<AbstractBuild>> futures = super.perform2(build, launcher, listener);
-        if(block==null) return ArrayListMultimap.create();
+    public ListMultimap<AbstractProject, QueueTaskFuture<AbstractBuild>> perform2(AbstractBuild<?, ?> build,
+            Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+        ListMultimap<AbstractProject, QueueTaskFuture<AbstractBuild>> futures = super.perform2(build, launcher,
+                listener);
+        if (block == null)
+            return ArrayListMultimap.create();
```
jglick, line 70: Again, best to just revert any change that does not further the stated goal of the PR.

jglick (commented):
Diff is much longer than it needs to be, but if the formatting changes were reverted I think this would be good.

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
dheerajodha: Yes sure, sorry about that. I'll undo the formatting changes. 
My VS Code did these automatically, I'll keep this in mind from now on.

jglick: Yeah some editors default to just reformatting everything on save. Fine for Go since everyone uses `gofmt` to begin with, but bad for diffs in other languages.

jglick:
https://github.com/jenkinsci/parameterized-trigger-plugin/pull/235/checks?check_run_id=5243312168 smells like a too-aggressive timeout?

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -74,7 +78,7 @@ protected QueueTaskFuture schedule(AbstractBuild<?, ?> build, Job project, List<
                 // which is most likely unintended. Might make sense to do it at BuildTriggerConfig for all.
                 list = CollectionUtils.immutableList(list, new DifferentiatingAction());
 
-                
+
```
jglick, line 81: FYI

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
```diff
@@ -96,4 +100,4 @@ protected QueueTaskFuture schedule(AbstractBuild<?, ?> build, Job project, List<
     @Extension
     public static class DescriptorImpl extends BuildTriggerConfig.DescriptorImpl {
     }
-}
+}
```
jglick, line 103: also

src/main/java/hudson/plugins/parameterizedtrigger/FileBuildTriggerConfig.java
```diff
@@ -19,6 +21,7 @@ public FileBuildTriggerConfig(String projects,
     private ResultCondition condition;
     private boolean triggerWithNoParameters;
     private boolean includeCurrentParameters;
+	@SuppressFBWarnings(value="UUF_UNUSED_FIELD", justification="Do not risk compatibility")
     private boolean batchCondition;
```
jglick, line 25: If it is in fact unused that it should be safe to delete?

src/main/java/hudson/plugins/parameterizedtrigger/PredefinedPropertiesBuildTriggerConfig.java
```diff
@@ -31,4 +31,4 @@ public Object readResolve() {
     	}
 		return new BuildTriggerConfig(projectsValue, condition, triggerWithNoParameters, configs);
     }
-}
+}
```
jglick, line 34: revert

jglick (approved):
Thanks!

src/main/java/hudson/plugins/parameterizedtrigger/BlockableBuildTriggerConfig.java
dheerajodha: 😅 thanks!

dheerajodha: Quick general question: 
Correct me if I'm wrong but ideally, there should be just one line after line 79, but currently, there are 2 empty lines. 
So, we don't want to correct it because we don't want to make any changes in the original code just for the sake of keeping things the way they are unless it's really urgent. Is this what our intention is?

dheerajodha: Understood :+1: 
also wow, the comments on the PR you shared above, are kind of intense. I get your point now.

dheerajodha:
> https://github.com/jenkinsci/parameterized-trigger-plugin/pull/235/checks?check_run_id=5243312168 smells like a too-aggressive timeout?

I tried to see what the error was:

> Condition with lambda expression in hudson.plugins.parameterizedtrigger.test.BuildTriggerTest that uses hudson.model.FreeStyleProject was not fulfilled within 10 seconds.

But now it is resolved, and all I did was undo the autoformatting done by IDE. I'm confused about what went wrong and how it got resolved...

jglick:
Probably just some flaky test unrelated to your PR.

src/main/java/hudson/plugins/parameterizedtrigger/FileBuildTriggerConfig.java
dheerajodha: Okay :+1: 
I tried to look for all the config.xml files within this repo and found one occurrence of this class in one of the config.xml [right here](https://github.com/jenkinsci/parameterized-trigger-plugin/blob/master/src/test/resources/hudson/plugins/parameterizedtrigger/test/CompatibilityTest/jobs/old/config.xml#L22). 

But based on your second sentence, then would it be okay to delete this field?

jglick: Oh perfect, so if you leave https://github.com/jenkinsci/parameterized-trigger-plugin/blob/eb2d40720a4b1e37ea38c1e01fdf91d485fc7e5e/src/test/resources/hudson/plugins/parameterizedtrigger/test/CompatibilityTest/jobs/old/config.xml#L24 alone and the corresponding `@LocalData` test still passes with the field deleted, that is some confirmation that it was safe.

src/main/java/hudson/plugins/parameterizedtrigger/FileBuildTriggerConfig.java
```diff
@@ -1,5 +1,7 @@
 package hudson.plugins.parameterizedtrigger;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
```
jglick, line 4:
```suggestion
```

pom.xml
```diff
@@ -217,7 +220,7 @@
     <changelist>-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/parameterized-trigger-plugin</gitHubRepo>
     <java.level>8</java.level>
-    <jenkins.version>2.270</jenkins.version>
+    <jenkins.version>2.289.1</jenkins.version>
```
olamy, line 223: as recommended here https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
can I change to `later releases within an LTS line`
thanks

olamy (dismissed):

pom.xml
jglick: The current PR matches the currently documented advice, but FWIW there is no real consensus here. See https://github.com/jenkins-infra/jenkins.io/pull/4876

src/main/java/hudson/plugins/parameterizedtrigger/Plugin.java
```diff
@@ -20,6 +22,7 @@
 
 import static org.apache.commons.lang.StringUtils.isEmpty;
 
+@SuppressFBWarnings(value="NM_SAME_SIMPLE_NAME_AS_SUPERCLASS", justification="Part of public API, reviewed all uses of Plugin are fully quantified in this class")
 public class Plugin extends hudson.Plugin {
```
jglick, line 26: Note that extending `Plugin` is deprecated. https://github.com/jenkinsci/parameterized-trigger-plugin/blob/5923a07ad166c3d77a4af0d3eab6d1a8c0305bdc/src/main/java/hudson/plugins/parameterizedtrigger/Plugin.java#L31-L32 _could_ be made into an `@Initializer`, though it could probably simply be a `static` initializer in `ResultCondition.java`.

jglick (approved):

src/main/java/hudson/plugins/parameterizedtrigger/Plugin.java
dheerajodha: I understood that extending `Plugin` is deprecated. 
Can you please help me in understanding what is our concern with `Stapler.CONVERT_UTILS.register` and how would that be addressed by making it into an `@Initializer`?

I read about `CONVERT_UTILS` from its [javadocs](https://javadoc.jenkins-ci.org/component/stapler/org/kohsuke/stapler/Stapler.html) and since ResultCondition is an enum, maybe we are trying to register it so that it becomes easier for Stapler to bind the enum constants...?

pom.xml
dheerajodha: Thanks for your comments! 
Lmk if we mutually decide to change `jenkins.version` be 2.289.3. I'm keeping it as 2.289.1 for now, unless Olivier has something to add.

src/main/java/hudson/plugins/parameterizedtrigger/Plugin.java
jglick: (To be clear, this is just an aside about an existing deprecated API, certainly not something that needs to be changed in this PR.)

dheerajodha: Thanks for the detailed info! This makes things little bit more clear to me now.

MarkEWaite (approved):

PR #240, opened 2022-01-30
Deeper spotbugs checks

dheerajodha:
Increase the chance of detecting bugs


src/main/java/hudson/plugins/parameterizedtrigger/SubProjectData.java
```diff
@@ -42,6 +44,7 @@
  *
  * @author JO Sivtoft
  */
+@SuppressFBWarnings(value="SIC_INNER_SHOULD_BE_STATIC_ANON")
```
dheerajodha, line 47: Should I add the "justification" field here inside the SuppressFBWarnings? If yes, then can anyone please tell me what exactly should I write here?

dheerajodha: Hi @jglick thank you for looking into this PR!
However, I'd need a couple of days to get back to this as I was (and am a little) preoccupied with job interviews + GSoC's application phase.

MarkEWaite (approved):

PR #266, opened 2022-07-08
[Flakiness Reduction] rebuild graph manually 

Pldi23:
https://issues.jenkins.io/browse/JENKINS-68964

Fix flickering BuildTriggerTest#testParentProjectTrigger

Building dependency graph is async process, and from time to time it is not complete when upstream job finished. 
Suggestion is to build graph manually as most of the tests in plugin already do.

### DIFF
--- a/src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerTest.java
+++ b/src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerTest.java
@@ -31,10 +31,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.*;
 
-import org.awaitility.Awaitility;
 import org.junit.Rule;
 import org.junit.Test;
 import hudson.tasks.Builder;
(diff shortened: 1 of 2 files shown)


jtnord:
> FYI: hudson.plugins.parameterizedtrigger.CapturedEnvironmentActionTest.onLoad failing on master also

as the test is trying to run `sh` just add an `@Assume.assumeFalse("test can not run on windows", Function.isWindows());` to the test.

can you link to the flake details in an OSS jira or inline here and also explain how this fixes the flake?

As I understand it - if the build has finished - then if the graph is still recomptint - recomputing the graph should have no affect (as the dependency is actually missing).  in other words shouldn;t rebuilding the graph need to take place *before* the build completes to avoid the flake (which would possibly be best done before the job was even started?)

src/test/java/hudson/plugins/parameterizedtrigger/test/BuildTriggerTest.java
```diff
@@ -70,7 +68,18 @@ public void testParentProjectTrigger() throws Exception {
...
-        Awaitility.await().pollInterval(1, SECONDS).atMost(10, SECONDS).until(() -> downstream.getLastBuild() != null);
+        r.waitUntilNoActivity();
+
+        if (downstream.getLastBuild() == null) {
+            /*
+             * There could be a period when upstream already finished, but async rebuilding graph {@link Jenkins#rebuildDependencyGraphAsync()}
+             * could still be in progress or even not start.
+             * Let's rebuild graph manually in that case to avoid flakiness.
+             */
+            r.jenkins.rebuildDependencyGraph();
```
jtnord, line 79: would it not be better to do this before the build was even started? and then you avoid the whole need for another build.  (which is what the tests in Jenkins core actually [appear to do](https://github.com/jenkinsci/jenkins/search?l=Java&q=rebuildDependencyGraph)?

are there other places in this plugins tests that also show the same issues that should be fixed also, or do they already rebuild the dependency graph correctly?

Pldi23: >would it not be better to do this before the build was even started? and then you avoid the whole need for another build. (which is what the tests in Jenkins core actually [appear to do](https://github.com/jenkinsci/jenkins/search?l=Java&q=rebuildDependencyGraph)?

Yes it's a good option, this is actually what most of the test already do. (changes in [3ea7129](https://github.com/jenkinsci/parameterized-trigger-plugin/pull/266/commits/3ea7129a9ec00173db196122552f9f4402e0896d))

>are there other places in this plugins tests that also show the same issues that should be fixed also, or do they already rebuild the dependency graph correctly?

Most of the test call `r.jenkins.rebuildDependencyGraph();` but not all of them, I will investigate if there could be something related.

Pldi23:
@jtnord could you please merge this (as I don't have permission)? 
And one more question would it be released automatically using maven-release plugin or we need to do it manually (if manually I'm not sure if I have permissions also)?

PR #291, opened 2022-09-29
[WIP] JENKINS-64591 Matrix Job - Trigger downstream Workflow job for all childs only when configured

gjabouley-invn:
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira: https://issues.jenkins.io/browse/JENKINS-64591
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

This issue only arises for Matrix Job.
When configuring a downstream job (post-build step), not enabling the option `Trigger build from child projects`, it is expected that only one run of the downstream job will be executed.
Turns out it is right for FreeStyle downstream job, but wrong for Workflow downstream job.

For a Matrix Job, it seems the downstream jobs will get triggered by 2 different mechanisms:
* dependency graph in case of FreeStyleJob
* direct call in case of Workflow job

Root cause of the issue is the UpstreamCause object is not created in the same "project level" context.
So in [constructor](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Cause.java#L180-L181) some calls will return different values:
* For FreeStyleJob, Matrix Parent job name (ie. upstream_job) and url
* For Workflow job, Matrix Parent configuration job name (ie. upstream_job/axis1=test) and url

Result is:
* For FreeStyleJob, all causes are considered "equals", so only one downstream job is started
* For Workflow job, each cause is considered unique (see [UpstreamCause.equals()](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Cause.java#L214-L226) implementation)

Proposed fix is to handle the case `build instanceof MatrixRun && ! triggerFromChildProjects` when creating the UpstreamCause, and in such case use the parentBuild() of the MatrixRun, so its name won't have the matrix configuration added (/axis1=test)

I have added 2 test cases to illustrate this behavior.

Sadly, it is breaking other tests from `FileBuildTriggerConfigTest`, and i am not sure how to fix this yet... Any help appreciated

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/BuildTriggerConfig.java
@@ -5,6 +5,7 @@ import com.google.common.collect.ListMultimap;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Launcher;
+import hudson.matrix.MatrixRun;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.Action;
@@ -529,6 +530,10 @@ public class BuildTriggerConfig implements Describable<BuildTriggerConfig> {
 
                 return new UpstreamCause((Run<?,?>)promotion.getTargetBuild());
             }
+        }
+        if(build instanceof MatrixRun && ! triggerFromChildProjects) {
+            MatrixRun matrixRun = (MatrixRun)build;
+            return new UpstreamCause((Run<?,?>)matrixRun.getParentBuild());
         }
         return new UpstreamCause(build);
     }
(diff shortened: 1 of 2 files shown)


gjabouley-invn:
Downstream jobs execution are managed differently in Jenkins:
- if a freestyle job, it is managed by DependencyGraph, and the plugin is behaving correctly and consistently
- if a workflow job, it is managed by the plugin itself, and in that case the following can be observed:
    - if option `triggerFromChildProjects` is activated, then a downstream execution of the workflow job is properly scheduled for each MatrixRun
    - if option `triggerFromChildProjects` is **disabled**, still a downstream execution of the workflow job is wrongly scheduled for each MatrixRun. I was not able to find a way to get the plugin function [perform](https://github.com/jenkinsci/parameterized-trigger-plugin/blob/master/src/main/java/hudson/plugins/parameterizedtrigger/BuildTrigger.java#L69) to be called for the Parent MatrixRun...

PR #375, opened 2024-04-10
Do not warn when triggered job is a success

hashar:
I keep having log messages such as:

  WARNING: [hudson.plugins.parameterizedtrigger.TriggerBuilder perform]
  Attempting to use the result of unfinished build <name of job> #1234

The message was introduced in #161 by 06d8f19218 but in #148 / 02ee6ea5, mapBuildResult() was made to return `null` by default (eg on SUCCESS).

The resut is the warning is always triggered when the triggered build succeeds.

Fixes https://phabricator.wikimedia.org/T336782

### Testing done

```[tasklist]
### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
@@ -187,11 +187,8 @@ public class TriggerBuilder extends Builder implements DependencyDeclarer {
                                     if (buildStepResult
                                             && config.getBlock().mapBuildStepResult(completedRun.getResult())) {
                                         Result r = config.getBlock().mapBuildResult(completedRun.getResult());
-                                        if (r != null) {
+                                        if (r != null) { // The blocking job is not a success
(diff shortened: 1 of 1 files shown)


MarkEWaite:
This pull request is a good reason to release a new version of the plugin, though I would like to complete the evaluation of the upgrade of the optional dependency on promoted builds plugin before this is released.  The optional dependency is being evaluated in two different pull requests:

* #378 
* #367 

My preference is to update to the older version of promoted builds so that the fewest users are forced to update to a newer version of promoted builds, but the dependency version is still modernized to something that is only 2 years old instead of something that is much, much older than that.

hashar:
Do you need me to rebase the pull request?

MarkEWaite:
Thanks for asking, but no need to rebase. It will be squash merged

PR #363, opened 2023-11-21
Cancel pending triggered build when parent project is cancelled

hashar:
Cancel triggered build on InterruptException

When a triggering project is cancelled by interrupting its Executor thread and a triggered project is waiting in the build queue (via `waitForStart()`) the InterruptedException is not handled.  The tiggered project is properly cancelled, but the triggering project is left in the queue and will eventually run when a slot is available.

Catch InterruptException while waiting for the triggered project to start and cancel the queued run.

Stacktrace with parameterized-trigger-2.44:

```
org.jenkinsci.plugins.postbuildscript.PostBuildScriptException: java.lang.InterruptedException
    at org.jenkinsci.plugins.postbuildscript.processor.Processor.processBuildSteps(Processor.java:190)
    at org.jenkinsci.plugins.postbuildscript.processor.Processor.processScripts(Processor.java:91)
    at org.jenkinsci.plugins.postbuildscript.processor.Processor.process(Processor.java:79)
    at org.jenkinsci.plugins.postbuildscript.processor.Processor.process(Processor.java:73)
    at org.jenkinsci.plugins.postbuildscript.PostBuildScript.perform(PostBuildScript.java:116)
    at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
    at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:803)
    at hudson.model.AbstractBuild$AbstractBuildExecution.performAllBuildSteps(AbstractBuild.java:752)
    at hudson.model.Build$BuildExecution.post2(Build.java:177)
    at hudson.model.AbstractBuild$AbstractBuildExecution.post(AbstractBuild.java:697)
    at hudson.model.Run.execute(Run.java:1932)
    at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
    at hudson.model.ResourceController.execute(ResourceController.java:97)
    at hudson.model.Executor.run(Executor.java:429)
Caused by: java.lang.InterruptedException
    at java.base/java.lang.Object.wait(Native Method)
    at java.base/java.lang.Object.wait(Object.java:328)
    at hudson.remoting.AsyncFutureImpl.get(AsyncFutureImpl.java:79)
    at hudson.model.queue.FutureImpl.waitForStart(FutureImpl.java:68)
    at hudson.plugins.parameterizedtrigger.TriggerBuilder.perform(TriggerBuilder.java:146)
[...]

PR #400, opened 2024-10-18
Handle NOT_BUILT and ABORTED as other results

hashar:
Use case
--------

My use case is a post build script triggering a job which does not need to fully complete. Since the build step can be cancelled while waiting or building, it causes the triggering build to be marked as a failure despite asking to never block.

To fix that, I need the plugin to pass the result of the build step through `BlockingBehaviour`. That lets one define how to behave when the triggered build is `NOT_BUILT` (it got cancelled frmo the queue). In my case I need it to never block.

See: https://phabricator.wikimedia.org/T352319

Solution
--------

There are multiple reasons for a job to not fully complete:

- it is interrupted, an InterruptedException is thrown, this is still rethrown and Jenkins will mark the build as `ABORTED`.

- it can be cancelled from the build queue raising a `CancellationException`. This previously raised an `AbortException` which Jenkins handles by marking the build as a failure. I have changed it to a `NOT_BUILT` result which can be process as other results (addressing  my use case to have it to never block).

The Jenkins Result class ranks the results as:
- `SUCCESS`
- `UNSTABLE`
- `FAILURE`
- `NOT_BUILT`
- `ABORTED`.

The `NOT_BUILT` and `ABORTED` results are thus worse than a `FAILURE` and would be matched as such in `BlockingBehavior` `mapBuildStepResult()` and `mapBuildResult()` which both use `isWorseOrEqualTo()` for comparison.

Add a test `testCancelledFromBuildQueue()` to cover the `CancellationException()` is caught and it results in a `SUCCESS` (since the test blocking behavior is to never block).

The `ResultConditionTest` test covers that `BlockingBehavior` is able to map `NOT_BUILD` and `ABORTED` since it has two tests explicitly cancelling and interrupting jobs.

Examples
--------

When a build is ongoing and when aborting it:
```
  Waiting for the completion of downstream-project
  downstream-project #7 started.
  downstream-project #7 completed. Result was ABORTED
  Build step 'Trigger/call builds on other projects' marked build as failure
  Finished: FAILURE
```

When it is waiting in the build queue and get cancelled:
```
  Waiting for the completion of downstream-project
  Not built: downstream-project has been cancelled while waiting in the queue.
  Build step 'Trigger/call builds on other projects' marked build as failure
  Finished: FAILURE
```

### Testing done

I have created two jobs in a Jenkins spinned up with `mvn hpi:run`. I have covered the behavior for cancelled jobs with a new test `TriggerBuilderTest.testCancelledFromBuildQueue()`.

The result mapping for `ABORTED` and `NOT_BUILD` is already covered by `ResultConditionTest`.

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch! _I have used a branch named `handle-not_built-canceled`_
- [x] Link to relevant issues in GitHub or Jira
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
+++ b/src/main/java/hudson/plugins/parameterizedtrigger/TriggerBuilder.java
@@ -151,52 +151,50 @@ public class TriggerBuilder extends Builder implements DependencyDeclarer {
                         }
                         for (QueueTaskFuture<AbstractBuild> future : futures.get(p)) {
                             try {
-                                if (future != null) {
-                                    listener.getLogger()
-                                            .println("Waiting for the completion of "
-                                                    + HyperlinkNote.encodeTo('/' + p.getUrl(), p.getFullDisplayName()));
-                                    Run startedRun;
-                                    try {
-                                        startedRun = future.waitForStart();
-                                    } catch (InterruptedException x) {
-                                        listener.getLogger()
-                                                .println("Build aborting: cancelling queued project "
-                                                        + HyperlinkNote.encodeTo(
-                                                                '/' + p.getUrl(), p.getFullDisplayName()));
-                                        future.cancel(true);
-                                        throw x; // rethrow so that the triggering project get flagged as cancelled
-                                    }
-
-                                    listener.getLogger()
-                                            .println(HyperlinkNote.encodeTo(
-                                                            '/' + startedRun.getUrl(), startedRun.getFullDisplayName())
-                                                    + " started.");
-
-                                    Run completedRun = future.get();
-                                    listener.getLogger()
-                                            .println(HyperlinkNote.encodeTo(
-                                                            '/' + completedRun.getUrl(),
-                                                            completedRun.getFullDisplayName())
-                                                    + " completed. Result was " + completedRun.getResult());
-                                    BuildInfoExporterAction.addBuildInfoExporterAction(
-                                            build,
-                                            completedRun.getParent().getFullName(),
-                                            completedRun.getNumber(),
-                                            completedRun.getResult());
-
-                                    if (buildStepResult
-                                            && config.getBlock().mapBuildStepResult(completedRun.getResult())) {
-                                        Result r = config.getBlock().mapBuildResult(completedRun.getResult());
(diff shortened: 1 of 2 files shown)


hashar:
In https://github.com/jenkinsci/parameterized-trigger-plugin/compare/9f7e52bbba881e1444aa22d8cbde18f7a5ed0a03..a95cb34ef00de439568ce662fbdc083624b15247 I did:
```(diff)
  } catch (InterruptedException x) {
  ...
- completedResult = Result.ABORTED
+ throw x; // rethrow so that the triggering project get flagged as cancelled
```

The commit message and the pull request text should be adjusted to take that in account. I have thus marked this as being in draft until I fix the texts.

hashar:
From https://phabricator.wikimedia.org/T352319#10288944

**Not built**

A build which was waiting for a triggered project which **never started**:
```
00:00:37.918 [PostBuildScript] - [INFO] Executing post build scripts.
00:00:37.920 Waiting for the completion of castor-save-workspace-cache
00:04:38.831 Not built: castor-save-workspace-cache has been cancelled while waiting in the queue.
```
That one got cancelled by receiving a `CancellationException` after four minutes. The option _Fail this build step if the triggered build is worse than or equal to_, which is set to `never`, **is now properly managed and ignored**.

**Build aborting**

An other build has:
```
00:01:45.403 Waiting for the completion of castor-save-workspace-cache
00:03:33.938 Build aborting: cancelling queued project castor-save-workspace-cache
```
That is the caller job being aborted by Zuul (our CI system driving Jenkins), the triggered job is aborted. _Fail this build step if the triggered build is worse than or equal_, to which is set to `never`, **ignores it**. That is catching the `InterruptedException` which I guess should be rethrown. I did that in https://github.com/jenkinsci/parameterized-trigger-plugin/pull/363/

hashar:
Hi @jenkinsci/parameterized-trigger-plugin-developers , I wrote this change a couple years ago and we had it deployed for sometime until an update of plugin eventually replaced the forked version I had deployed.

Would it be possible for one of you to have a look at it?  It is hopefully thoroughly detailed. Thank you :-]

